PR #121, opened 2012-11-30
support for AG_OPTIONS

gurgeous:
Please be gentle - I haven't written any C in a while. Feel free to edit. :)

I renamed the original argc/argv to argc0/argv0.

I didn't bother freeing the copies at the end. I noticed some other leaks with valgrind, so I figured you don't care. Because of the way argc & argv are used you'd have to make a copy of the original pointers if you want to free them later.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -1,3 +1,4 @@
+#include <ctype.h>
 #include <errno.h>
 #include <stdarg.h>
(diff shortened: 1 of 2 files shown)


gurgeous:
I think it's time to close this guy :)

PR #202, opened 2013-04-30
Defined wordchar table

afredd:
Explicitly define the array wordchar_table[].
- use unsigned char and not int - minor space saving
- smaller code footprint (no init_wordchar_table() code)

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -89,7 +89,6 @@ int main(int argc, char **argv) {
         }
         generate_skip_lookup(opts.query, opts.query_len, skip_lookup, opts.casing == CASE_SENSITIVE);
         if (opts.word_regexp) {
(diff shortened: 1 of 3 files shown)


forgottenswitch:
From http://stackoverflow.com/questions/10670213/efficiency-char-array-vs-int-array:
```
Some CPUs can access bytes without problems.
If you can throw SIMD at it, char is likely to win.
```

So `int -> char` isn't worth it.
And is the no-`init_wordchar_table` ?

ggreer:
Sorry, but this is harder to understand, more lines of code, and less likely to work across platforms. There's a slight performance improvement since `wordchar_table[]` is already initialized, but it's not worth the extra complexity and brittleness.

PR #228, opened 2013-06-28
prints files and numbers like './build.md:70'

CarlQLange:
I wanted to be able to list files and numbers of matches so I could do a checklist-type-thing. So now, when you do `ag -N "myMatch"`, (or `ag --files-and-numbers "myMatch"`) it prints out a list of matching files like 

```
file1.cpp:42
file1.cpp:57

file2.cpp:29
```

and you can use it with `--nobreak` to get rid of the newlines if you don't like them. It has coloured and uncoloured output and stuff too. :yellow_heart:

By the way, great code! Haven't written C in a _long_ time, the source was very legible and extensible. Let me know if I missed or broke anything!


CarlQLange:
closing since I never got around to this and it's many many years later :)

PR #295, opened 2013-11-07
New option --only-matching.

KingDuckZ:
New grep-like option to only display the text that matched the input regex.


neojski:
This is already implemented, see [docs](https://github.com/ggreer/the_silver_searcher/blob/8cc74e04eadce843317bece2feb73d36d02d7529/doc/ag.1.md).

@KingDuckZ or @ggreer, you can guys close this one :-)

PR #332, opened 2014-01-29
Add option --list-file-types

mcphail:
This patch documents the file types which were made searchable in #283


kamal94:
Please resurrect this. I dont know what file types are supported.

PR #333, opened 2014-02-05
Support color on windows

mattn:


ggreer:
Thanks for keeping this up to date so I could merge it.

When you first created this PR, I saw "windows" and "+300 lines" and thought, "Uh oh, there's no way I can merge this." I avoided even looking at the diff. Today I finally noticed that you kept everything contained in a separate file. Guess I should have looked at this sooner. Oops.

mattn:
:+1:

PR #356, opened 2014-03-20
[WIP] Wildmatch

kans:
The goal is to parse gitignore regexes the way git does it (using git's wildmatch.c).


RedX2501:
So i have asked on the git mailing list about getting wildmatch relicensed and am waiting for a reply. See #614 for the status.

PR #427, opened 2014-06-12
Added ".bzrignore" to list of VCS ignore files

PhilHudson:
Adds support for the Bazaar DVCS.


ggreer:
Since every entry in `ignore_pattern_files` is another call to `fopen()` in every directory ag looks at, I'm not too keen on adding more to it. Also, [development on bazaar has stalled](https://www.openhub.net/p/bazaar/commits/summary). There hasn't been a release in over two years. New projects aren't choosing it and old projects are moving off it.

In other words, the performance penalty is paid by everyone, but the feature benefits only a few. IMO, that trade-off isn't worthwhile. Sorry.

PhilHudson:
On Sun, 14 Feb 2016 at  7:15:59 am GMT, Geoff Greer <notifications@github.com> wrote:

> Since every entry in ignore_pattern_files is another call to fopen() in every
> directory ag looks at, I'm not too keen on adding more to it. Also, development
> on bazaar has stalled. There hasn't been a release in over two years. New
> projects aren't choosing it and old projects are moving off it.
>
> In other words, the performance penalty is paid by everyone, but the feature
> benefits only a few. IMO, that trade-off isn't worthwhile. Sorry.

Fair enough. Thanks for looking at it.

-- 
Phil Hudson                   http://hudson-it.ddns.net
@UWascalWabbit                 PGP/GnuPG ID: 0x887DCA63

PR #479, opened 2014-08-24
Stdin fixes

### LINKED ISSUE #476: ag only detects pipes but not simple input redirection
ag detects correctly that it should read from STDIN if I pipe something through it, but not if I use simple redirection from a file. Observe:

```
[0 mbunkus@chai-latte ~/tmp/empty] ls
[0 mbunkus@chai-latte ~/tmp/empty] cat ../samba.txt | ag just
just
[0 mbunkus@chai-latte ~/tmp/empty] ag just < ../samba.txt
[1 mbunkus@chai-latte ~/tmp/empty]
```

luksen:
Mainly, this fixes #476 (further comments in commit). The other two I stumbled upon while reading the relevant code.

The solution with `/dev/stdin` works nicely on my system, but I'm not sure how portable it (or ag for that matter) is.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -633,7 +633,24 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
         /* Make sure we search these paths instead of stdin. */
         opts.search_stream = 0;
     } else {
-        path = ag_strdup(".");
+        /* If stdin is a regular file, try to open it as portably as possible.
+         * If none of the filesystem mappings for stdin exist, fall back to
+         * treating it as a stream. If stdin isn't regular, no file was
+         * specified by the user, so search the current directory.
+         */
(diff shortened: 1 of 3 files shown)


ggreer:
`--parallel` isn't used in the passthrough test. The passthrough test setup unaliases `ag` and re-aliases it with `--nocolor --workers=1`.

I'm sorry your PR didn't get merged, but thank you for the effort. As a consolation prize, I cherry-picked your goto commit.

luksen:
As I've said, I'm fine with this being closed. But just to clarify, the point of my September comment was that the `--parallel` in the other tests masks the imho faulty behaviour. The passthrough test shows it because it doesn't use `--parallel` and I believe my original testcase from the passthrough PR is faulty for allowing it. I guess I was trying to sneak in a fix for my own mistake ;)

PR #487, opened 2014-09-03
Fix literal patterns containing slashes.

eevee:
This fixes both `--ignore foo/bar` on the command line and `foo/bar` inside a VCS ignore file.  Both of which I've tripped over.  :)


eevee:
Closing this, since it's been fixed in the meantime — the tests I wrote pass under ag 0.31.0.

PR #512, opened 2014-10-16
Add -x and --files-from options.

sruffell:
Take 2: Not sure if you're interested but these options mimic the 'ack' options which enable using another program to generate the list of paths to search.

I had a typo in the first pull request where the paths and base_paths arguments to the add_paths function was reversed, but I also added a test for the files-from option specifically.


sruffell:
I just noticed that this is still open and it's probably pretty stale at this point. I figure I'll just close it out.

PR #542, opened 2014-11-26
fix multiple file types issue

alexandnpu:
Originally, when there are multiple file types presented, ag only takes
the last one as its file type.
But the user want to search all the file types.
e.g. when i type `ag --cpp --cc what_i_want`, ag only searches for c
files.
But the user wants to seach both c and cpp files.

This change fixes this.

### DIFF
--- a/src/lang.c
+++ b/src/lang.c
@@ -79,17 +79,19 @@ unsigned int get_lang_count() {
     return sizeof(langs) / sizeof(lang_spec_t);
 }
 
-char *make_lang_regex(const char **extensions) {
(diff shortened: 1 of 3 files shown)


ggreer:
Tests would be nice, but I've neglected this PR long enough that any changes are on me, not you. [My real work](https://floobits.com/) sucks up most of my coding time.

Thank you for writing this. I'll merge it and clean it up myself.

alexandnpu:
Thanks for merging. I am glad I can help.

PR #547, opened 2014-12-02
Use fnmatch on full path

blueyed:
Followup to #487, based on master.


blueyed:
According to https://github.com/ggreer/the_silver_searcher/pull/487#issuecomment-215927862 it should be fixed.
Closing.

PR #567, opened 2014-12-30
Turn off colors with -v. Plus a test for --color, --nocolor and -v. Fixes #478.

### LINKED ISSUE #478: highlighting for -v is distracting
The highlighting is a bit weird would be my only feedback.

``` bash
$ echo -e "blah foo blah\nfoo bar" > out.txt
$ ag bar out.txt | ag -v blah
2:foo bar
```

The whole foo bar is highlighted. It's easier to see if nothing is highlighted since usually the grep result from -v results in a lot of highlighted text.

screen shot 2014-08-24 at 11 27 22 am

gurgeous:


ggreer:
Perfect.

PR #573, opened 2015-01-06
Update README.md

morenoh149:
Added example usage


ggreer:
Sorry, but I think the readme is big enough, and that usage of ag is pretty straightforward. If someone really has trouble, they can always run `man ag` or `ag --help`.

PR #577, opened 2015-01-11
Add option to skip holes in sparse files.

umanwizard:
Hi and thanks for making "ag", one of the most invaluable tools that I use daily!

Previously "ag -u asdfasdfasdfadsf /root_partition" was taking nearly ten minutes, due to gobblying gigabytes of zeros in sparse files, notably /var/log/lastlog, which is nominally 400+ GB on my system but actually occupies less than 100 KB

The behavior witnessed is the worker thread pegging its core to 100%, as well as the kernel swapper thread getting its core to 40% dutifully discarding those pages of zeros as ag faults in ever more of them.

This adds a --skip-holes option that jumps over holes in sparse files, using the SEEK_HOLE and SEEK_DATA API described here: https://blogs.oracle.com/bonwick/entry/seek_hole_and_seek_data (and in the years since, implemented on more OSes than Solaris)

Now with --skip-holes I can do ag -u of my root partition in something like 1:30 instead of 9:30, and ag -u of /var/log/lastlog takes seconds.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -44,6 +44,8 @@ AS_IF([test "x$enable_lzma" != "xno"], [
     PKG_CHECK_MODULES([LZMA], [liblzma])
 ])
 
+AC_CHECK_DECL([SEEK_HOLE], [AC_DEFINE([USE_SEEK_HOLE], [], [Enable seeking to hole/data sections of files])], [], [#include <unistd.h>])
+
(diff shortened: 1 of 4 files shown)


umanwizard:
oops, closed this in favor of https://github.com/ggreer/the_silver_searcher/pull/579

PR #580, opened 2015-01-11
Add support for a --one-device flag

umanwizard:
Add a flag that will cause `ag` not to descend into mount points.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -157,7 +157,14 @@ int main(int argc, char **argv) {
             log_debug("searching path %s for %s", paths[i], opts.query);
             symhash = NULL;
             ignores *ig = init_ignore(root_ignores, "", 0);
-            search_dir(ig, base_paths[i], paths[i], 0);
+            struct stat s = {.st_dev = 0};
+            /* The device is ignored if opts.one_dev is false, so it's fine
+             * to leave it at the default 0
+             */
+            if (opts.one_dev && lstat(paths[i], &s) == -1) {
+                log_err("Failed to get device information for path %s. Skipping...", paths[i]);
+            }
+            search_dir(ig, base_paths[i], paths[i], 0, s.st_dev);
             cleanup_ignore(ig);
         }
         pthread_mutex_lock(&work_queue_mtx);
--- a/src/options.c
+++ b/src/options.c
@@ -126,7 +126,6 @@ void init_options(void) {
     opts.color_match = ag_strdup(color_match);
     opts.color_line_number = ag_strdup(color_line_number);
     opts.use_thread_affinity = TRUE;
-    opts.one_dev = FALSE;
 }
 
 void cleanup_options(void) {
--- a/src/options.c
+++ b/src/options.c
@@ -77,6 +77,7 @@ Search Options:\n\
                           (literal file/directory names also allowed)\n\
      --ignore-dir NAME    Alias for --ignore for compatibility with ack.\n\
   -m --max-count NUM      Skip the rest of a file after NUM matches (Default: 10,000)\n\
+     --one-device         Don't follow links to other devices.\n\
   -p --path-to-agignore STRING\n\
                           Use .agignore file at STRING\n\
   -Q --literal            Don't parse PATTERN as a regular expression\n\
@@ -125,6 +126,7 @@ void init_options(void) {
     opts.color_match = ag_strdup(color_match);
     opts.color_line_number = ag_strdup(color_line_number);
     opts.use_thread_affinity = TRUE;
+    opts.one_dev = FALSE;
 }
 
 void cleanup_options(void) {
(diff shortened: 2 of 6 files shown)


src/search.c
```diff
@@ -445,6 +446,17 @@ void search_dir(ignores *ig, const char *base_path, const char *path, const int
         queue_item = NULL;
         dir = dir_list[i];
         ag_asprintf(&dir_full_path, "%s/%s", path, dir->d_name);
+        if (opts.one_dev) {
+            struct stat s;
+            if (lstat(dir_full_path, &s) != 0) {
```
ggreer: I think you should put this block in `search_file()`. There's already a call to `stat()` in that function, so it would improve performance by avoiding a second call. You could also avoid passing `original_dev` around by putting it on `opts`. Since `original_dev` is read-only after main has set it, there would be no need to put a mutex lock around accesses to it.

src/options.c
```diff
@@ -125,6 +126,7 @@ void init_options(void) {
     opts.color_match = ag_strdup(color_match);
     opts.color_line_number = ag_strdup(color_line_number);
     opts.use_thread_affinity = TRUE;
+    opts.one_dev = FALSE;
```
ggreer: The whole struct is zeroed-out by `memset()`. There's no need for the explicit init in this case.

src/main.c
```diff
@@ -157,7 +157,14 @@ int main(int argc, char **argv) {
             log_debug("searching path %s for %s", paths[i], opts.query);
             symhash = NULL;
             ignores *ig = init_ignore(root_ignores, "", 0);
-            search_dir(ig, base_paths[i], paths[i], 0);
+            struct stat s = {.st_dev = 0};
+            /* The device is ignored if opts.one_dev is false, so it's fine
+             * to leave it at the default 0
+             */
+            if (opts.one_dev && lstat(paths[i], &s) == -1) {
+                log_err("Failed to get device information for path %s. Skipping...", paths[i]);
```
ggreer: Maybe this should abort the search? I'm not sure. It seems like a pretty bad thing if someone specifies `--one-device` and we can't figure out the device.

ggreer:
I left some comments. Sorry for the delay in reviewing your PR. Busy busy busy.

Please don't take any of my criticism personally. I want this feature in ag, I just think a little cleanup is necessary before merging.

src/main.c
umanwizard: `ag needle foo bar` will continue searching `bar` even if it can't open `foo`. Now similarly `ag --one-device needle foo bar` will continue searching `bar` if it can't open `foo`, or if it can't determine its device. I think emitting an error message and continuing with the rest of the search paths is the right behavior here, for consistency.

src/options.c
umanwizard: Good call, will fix.

src/search.c
umanwizard: I think that after the first call to `stat(2)`, the inode is in the buffer cache, so calls to `stat(2)` in the near future will not meaningfully I/O performance. And even though syscalls aren't free, I really doubt this will be a factor even in CPU-bound workloads. There might be value in doing some profiling just to make sure this is true, though.

If we put this check in `search_file` instead of `search_dir`, then we are still invoking non-trivial operations on another device, which one would expect `--one-device` to avoid. E.g., imagine if the mount point is a network FS over a slow link, with 1000 files in the top level of the FS. Iterating over the entries in the mount directory and `stat(2)`-ing each one, then skipping them all because they aren't on the same device, would take a long time.

A more usual case where this would cause bad behavior would be, e.g., descending into `/dev` and getting a bunch of annoying permissions errors when we try to `open(2)` a file (src/search.c:193) before ever getting the chance to `stat(2)` it.

Finally, imagine filesystem A has a mount point for filesystem B, within which there is a symlink that goes back to somewhere on A. As a user I would expect that symlink **not** to be followed, since `ag --one-device` "should" semantically never be looking into the mount point for B at all.

umanwizard: As for saving `original_dev` in `opts`, this value can be different for each path specified on the command line, so we would have to use some sort of array, or path-to-device hashmap. I think passing it to the function is simpler.

ggreer: Ah, that's right. I forgot ag lets you search multiple paths at once.

You make good points. I'm convinced. No change needed here.

ggreer:
One last thing: Do you have any idea how to write a test case for this code? I could totally see a regression because someone (probably me) forgot about this feature.

src/main.c
ggreer: Yeah. Totally forgot ag can search multiple paths. Current code seems fine then.

umanwizard:
No great options for testing come to mind...

On recent versions of Linux, /dev/shm is world-writable and almost certain to be a different device from wherever you're running the test from, so a test case could make a file there, symlink to it from the test directory, and verify that it's not read. But this is Linux-specific; it doesn't exist on FreeBSD (I just checked) and I have no idea about OSX.

I posted a question to Stack Overflow about this: http://stackoverflow.com/questions/28016940/is-there-a-reliable-cross-platform-on-nix-anyway-way-to-create-a-temporary-f . Maybe there is a way that I'm just not thinking of.

ggreer:
Cross-platform for that is too much to ask. It's fine if the test is linux-specific. I mostly work on a mac, but I test on linux machines and the Travis-CI stuff runs on Ubuntu. That should be enough of a gatekeeper to prevent regressions.

I would be super-happy if you add a /dev/shm test case. If you think it'll take a while, I'll just merge this PR for now.

umanwizard:
Done. Works on my Ubuntu 12.04 machine and, apparently, on your Travis CI build server as well.

ggreer:
:heart_eyes:

ggreer:
Thanks so much for being patient and responding to my comments. I'll likely tag a new release this week, and your feature will be in it.

PR #581, opened 2015-01-12
Fixed print.c terrible hack

r0nk:
This should fix the terrible hack in print.c


ggreer:
I guess a `size_t` initialized to `INT_MAX` is about the safest thing to do in this case. I think it can still break if `opts.after` is greater than `INT_MAX` and there are 2 billion lines in a file before the first match. Still, that's way better than the current behavior.

Thanks for the patch.

Issue #572, opened 2015-01-05
Searching for "||=" results in ag being OOM killed


YorickPeterse:
When searching for `||=` ag ends up being OOM killed by the kernel. A simple example:

```
mkdir /tmp/foo
cd /tmp/foo
echo 'foo ||= bar' > test.txt
ag '||=' .
```

This occurs at least on Arch Linux (`Linux tuxbook 3.17.3-1-ARCH #1 SMP PREEMPT Fri Nov 14 23:13:48 CET 2014 x86_64 GNU/Linux` in case it helps) using ag version 0.28.0.

umanwizard:
The issue here is that zero-length matches put Ag into an infinite loop.

ggreer:
Thanks to @umanwizard, this is fixed. A zero-length match is rather confusing when it's printed out, but who am I to second-guess someone's regex?

umanwizard:
FYI @YorickPeterse , `ag ||=` probably doesn't do what you want, if you are expecting it to find that literal string in files. You can pass the `-Q` flag to enable literal searching, rather than interpreting `||=` as a (very strange) regex.

YorickPeterse:
@umanwizard Hm, you're right, it didn't occur to me this would search for it as a regex. Thanks for looking into this!

PR #582, opened 2015-01-13
Fix zero length matches

umanwizard:
This fixes issue https://github.com/ggreer/the_silver_searcher/issues/572 , which is caused by an infinite loop when a zero-length match is encountered. Basically, when a zero-length match is found, the input buffer is not advanced, so the same match is found again, over and over. Each time, a new match object is created, eventually filling main memory and causing the program to be killed.

Normally Perl would not match the same zero-length match twice in a row, so this problem is avoided there. But that information is not tracked across calls to `pcre_exec`, so we need to advance the input to prevent it from happening.

An alternative fix would be to allow `pcre_exec` to return multiple matches. I believe (though I haven't tested it) that it would e.g. ignore the first zero-length match, and return the next possible match as the second match. However, this would become complicated if there is more than one possible zero-length match (e.g. if the regex is something like `^|(|a)`). More investigation would be required to determine the behavior of `pcre_exec` here, and design a solution that takes this possibility into account. Thus, I believe the simpler solution of just advancing the input is better.

Some background: http://www.regular-expressions.info/zerolength.html

### DIFF
--- a/src/print.c
+++ b/src/print.c
@@ -91,12 +91,6 @@ void print_file_matches(const char *path, const char *buf, const size_t buf_len,
     context_prev_lines = ag_calloc(sizeof(char *), (opts.before + 1));
 
     for (i = 0; i <= buf_len && (cur_match < matches_len || lines_since_last_match <= opts.after); i++) {
-        if (cur_match < matches_len && i == matches[cur_match].end) {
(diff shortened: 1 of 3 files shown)


ggreer:
Nice catch. Thank you so much for finding and fixing this.

Issue #574, opened 2015-01-08
Default set of options


sgraham:
Would it possible to set an envvar with preferred standard options? Or a config file?

I dislike grouping, so have to type `ag --nogroup` every time. I'm getting pretty good at it, but it destroys the careful optimization from strlen("grep") to strlen("ag") that I would otherwise experience.

MrDetonia:
Possibly by having ag check for an 'AG_OPTIONS' (or something similar) on startup, apply those options, and then handle command-line options afterwards, overwriting any settings in the environment variable.
'AG_OPTIONS' would take the form of space separated options: e.g. "nogroup nocolor column"
I imagine a config file would be a bit overkill for this alone.

This solution work for people?

sgraham:
Sounds perfect!

dighev:
You may want to see https://github.com/ggreer/the_silver_searcher/issues/111

MrDetonia:
My implementation for environment variables is working, pull request submitted (see above).
#111 mentions implementing an ~/.agrc config file or simply using aliases (although these can be a hassle). That too could serve a similar functionality.

Which solution would be preferred?
Why not both?

MrDetonia:
Stumped for the moment... any more that 3 program arguments (program name, one option, and a string to search for) causes a segmentation fault.
Anyone have any ideas why?

aswild:
Reviving this issue to present my own implementation of this feature: #1037

Issue #111, opened 2012-11-19
how create a preference file?


slimane:
one question, how create a .ackrc style preference file?

trinitronx:
:+1: for this feature...  People coming from `ack` seem to expect a portable way to use their `.ackrc` with `ag`, or at least modify & adapt their existing one for use with `ag`.  Aliases are ok for setting default arguments like `alias ag='ag --pager="less -SMFRX"'`, but things can get unwieldy pretty quick if you have a lot of arguments you want to add.  I can see that implementing this in a way that has to parse the file would be not very fast and would make it harder to maintain.

Perhaps the implementation could be quick and dirty:   simply check for an `~/.agrc` file, then if it exists read it and append the lines found to the `argv` array, then continue as normal.   This would let `parse_options(argc, argv, &base_paths, &paths);` do all the hard work, makes it easier to maintain and allows for future-proof support for other options added later.  If the `~/.agrc` file doesn't exist, then skip the whole code branch to read it in for speed.

diego898:
I'd just like to say this would be very useful for me. Its a shame that the only way to do this is with aliases.

PR #583, opened 2015-01-14
Default options via environment variable

MrDetonia:
ag can now read options from an environment variable 'AG_OPTIONS' as well as command line parameters.
This can be used to set default values for calls to ag, without the need for aliases or additional config files.

Implemented as a response to Issue #574

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -157,6 +157,88 @@ void cleanup_options(void) {
     }
 }
 
+void free_str_array(int len, char ***arr)
+{
+    int i;
+    for(i = 0; i < len; i++)
+    {
+        free(*(*arr+i));
+    }
+    free(*arr);
+}
+
+int append_env_options(int argc, char **argv, char ***all_opts)
+{
+    /*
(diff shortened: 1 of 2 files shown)


MrDetonia:
Issue seems to be that I've somehow broken optarg in the getopt_long loop in parse_options.
With the option --workers=1, optarg is 0x00, and atoi(optarg) is causing a segmentation fault.

nhooey:
@MrDetonia Could you give an example of how to set this command-line option in `AG_OPTIONS`?

```bash
ag --color-match='27;44'
```

MrDetonia:
@nhooey 
The exact arguments would be added to the environment variable, as though they were being appended to the command you type.

e.g. in your case:
````
export $AG_OPTIONS="--color-match='27;44'"
ag
````
should be equivalent to
````
ag --color-match='27;44'
````

nhooey:
Did this change not make it in to @ggreer's version 0.29.1?

MrDetonia:
The only other current method of doing something like this is using aliases ( see #111 )

PR #585, opened 2015-01-15
Default options via environment variable

MrDetonia:
Possibly fixed issue with +3 program arguments causing a segmentation fault.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -157,6 +157,88 @@ void cleanup_options(void) {
     }
 }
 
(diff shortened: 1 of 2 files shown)


MrDetonia:
For some reason, any more that 3 arguments to the program causes a segmentation fault.
I'm stumped for now. Any ideas anyone?

docwhat:
@MrDetonia I didn't trace all the `argv` usage path in your pull request so this is just a possibility: If you try to modify the `argv` pointer then bad things happen.

PR #586, opened 2015-01-17
Faster extension ignores

### LINKED ISSUE #584: Optimization: Ignoring file extensions
Right now, simple postfix patterns such as `*.txt` are loaded into ignore patterns and `fnmatch()`ed like everything else. It's probably worthwhile to detect these and load them into a sorted list. Then instead of calling `fnmatch()` on every pattern, we can binary search them.

I'll probably mess around with this optimization this weekend.

ggreer:
Fixes #584.


ggreer:
In my benchmarks, these changes are 10% slower. Apparently there aren't enough extensions (or they don't match often enough) to make binary searching worthwhile.

I still think it makes sense to optimize for this case, but I need to reconsider how to do it.

umanwizard:
Can you publish the benchmark you used? I'm interested in taking a look.

ggreer:
I just do stuff like time runs of searching my `~/code/` dir, which is 9.5GB.

    (ag)ggreer::ag::ftty ggreer@carbon:~/code% time ./ag/ag cpu_set_t
    ...
    ./ag/ag cpu_set_t  6.08s user 24.51s system 113% cpu 26.991 total


    (ag)ggreer::ag::ftty ggreer@carbon:~/code% time ag cpu_set_t
    ...
    ag cpu_set_t  8.60s user 24.12s system 124% cpu 26.309 total

To figure out what's taking up all the time, I use Instruments.app on my mac, or gprof on linux.

ggreer:
Now it's the same speed or faster. Hooray profiling.

ilyapopov:
Looks like there is a bug for patterns like `*.tar.gz`?

PR #593, opened 2015-01-21
Update RHEL7+ install instructions

etdev:
-Fix the RHEL7+ EPEL repository link (previous link is dead)


ggreer:
Thanks for this. Is there a canonical URL that won't change, or will RHEL require constant updates to the README?

PR #596, opened 2015-01-28
Fix Windows build.

living180:
Be sure to link against libshwlapi on Windows.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -27,6 +27,11 @@ AX_PTHREAD(
 CFLAGS="$CFLAGS $PTHREAD_CFLAGS $PCRE_CFLAGS -Wall -Wextra -Wformat=2 -Wno-format-nonliteral -Wshadow -Wpointer-arith -Wcast-qual -Wmissing-prototypes -Wno-missing-braces -std=gnu89 -D_GNU_SOURCE -O2"
 LDFLAGS="$LDFLAGS"
 
+case $host in
+*mingw*)
+    AC_CHECK_LIB(shlwapi, main,, AC_MSG_ERROR(libshlwapi missing))
+esac
+
 LIBS="$PTHREAD_LIBS $LIBS"
 
 AC_ARG_ENABLE([zlib],


ggreer:
Thanks for the patch. I hardly ever even use Windows, so I rely on people like you to fix any accidental breakage. :)

It looks like the only function needed from libshlwapi is `PathMatchSpec()`. Also, it's probably a good idea to check for the header file. Perhaps a better patch would be:

    AC_CHECK_HEADERS([shlwapi.h])
    AC_SEARCH_LIBS([PathMatchSpec], [shlwapi], [], AC_MSG_ERROR([libshlwapi missing]))

If you make those changes to your PR or tell me why your patch is better, I'll merge.

living180:
I'm by no means an autotools expert, so what I added was based on examples I was able to find in other projects.  The reason that checking for `PathMatchSpec` doesn't work is that it is a `__stdcall` function, so it gets mangled to the the symbol `_PathMatchSpec@8`, whereas `AC_SEARCH_LIBS` tries to link against it as a `__cdecl` function, which causes the linker to look for the symbol `_PathMatchSpec`, which does not exist.  The examples I found just checked for the function `main`, so that is what I did.

`AC_SEARCH_LIBS` doesn't work because if you check for `main`, it will succeed without needing to link in any libraries, and thus won't add `-llibshlwapi` to `LIBS` (not having `-llibshlwapi` in `LIBS` was the impetus for me creating this PR in the first place).

Finally, I don't see a good reason for the `AC_CHECK_HEADERS` test; `shwlapi.h` is provided by the same MinGW package as `libshlwapi.a`.  So if that package is missing, it will be caught by the `AC_CHECK_LIB` test.  In addition, the test as written above doesn't actually have any affect; it just defines `HAVE_SHLWAPI_H` if it finds the header and does nothing if it doesn't find it.  To be useful, it would need an `AC_MESSAGE_ERROR` to error out if `shlwapi.h` is not present.

It might be possible to write a more robust check using `AC_LINK_IFELSE` in combination with `AC_PROGRAM` but I don't think the work required would be worth the gain over what I have in this PR (it properly causes `-llibshlwapi` to be added to `LIBS`, and will cause `./configure` to fail in the unlikely case that someone is using MinGW without w32api support).

ggreer:
Those are good reasons. Thanks for spelling them out.

PR #597, opened 2015-01-31
Smarter word boundaries.

### LINKED ISSUE #563: Matching whole words fails in some cases
ag finds these matches as expected:

```
↪ ag -w '\.deliver'
spec/mailers/notification_mailer_spec.rb
41:      spam_notification_email.deliver
100:      failure_email.deliver
152:      route_limit_email.deliver
```

I'm using word match so it doesn't match on strings like `.deliveries`.
I then changed those instances of `.deliver` to `.deliver_now`.

Now ag doesn't find any more matches:

```
↪ ag -w '\.deliver'
```

But git grep finds remaining matches as expected:

```
↪ git grep -w '\.deliver'
app/controllers/webhooks/mailgun_controller.rb:      NotificationMailer.route_limit_notification(num_routes).deliver
app/controllers/webhooks/mailgun_controller.rb:    NotificationMailer.failed_delivery(params_and_message_headers).deliver
app/models/inbound_email.rb:      NotificationMailer.spam_notification(self).deliver
```

It seems that ag is failing to find matches when there's a `)` before the pattern I'm searching for.

pbhandari:
This fixes #563, where `-w` wasn't properly finding word boundaries.

All the tests pass still, I did my own testing on the command line and that seems to pass, if this PR seems OK to you guys i'll add a test file for this as well.

My issue here is that in order to have this fix work on literal(`-Q`) queries, I needed to change the definition of what counts as a wordchar to any _non-wordchar_ regardless of whether it currently follows a _wordchar_ or not.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -97,6 +97,10 @@ int main(int argc, char **argv) {
         opts.casing = is_lowercase(opts.query) ? CASE_INSENSITIVE : CASE_SENSITIVE;
     }
 
+    if (opts.word_regexp) {
+        init_wordchar_table();
+    }
+
     if (opts.literal) {
         if (opts.casing == CASE_INSENSITIVE) {
             /* Search routine needs the query to be lowercase */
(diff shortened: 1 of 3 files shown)


okdana:
This is a really neat change, but there is one problem with it, if i'm not mistaken — the handling of PCRE patterns is naïve, in that it treats characters in the pattern as literal, individual characters rather than as potential regular-expression atoms/meta-characters.

For example, in the pattern `\d{3}`, a naïve check of the beginning and end of the string will find no reason to add surrounding `\b`s because `\` and `}` are non-word characters — when actually the pattern matches nothing *but* word characters. There are a million other examples i can think of that this would cause problems with but i bet you get the idea.

Ack deals with word boundaries similarly, which i mentioned favourably in #563, but i wasn't thinking at the time that it's got this same flaw. It's actually worse than the original issue too — at least the current behaviour is consistent. With this method you would sometimes get no boundary checks at all, just depending on the pattern.

FreeBSD and GNU grep both seem to handle this by trying to match the basic pattern *first* and only then checking to see if the match exists between word boundaries.

pbhandari:
Damn, you're right, I hadn't considered that possibility.  Shouldn't be too hard of a fix though.

pbhandari:
Hopefully, this should fix it.  I'll get the tests up as soon as I can.

But again, going under the assumption that a `non-wordchar` is always a word boundary, instead of the old 'non-wordchar|wordchar`

okdana:
This is very similar to the way `grep` does it, and it works great for me. Thanks!

pbhandari:
Np, I needed this too, and now at-least it works properly :)

PR #610, opened 2015-02-19
Use fwrite to print lines

artygus:
Printing search results with --vimgrep is slow as heck when there're plenty of search matches. This little patch speeds up printing of search results a little bit. It gave me 3x boost in a project with around 3,000 files and short search queries


ggreer:
Nice catch.

PR #612, opened 2015-02-20
Use strrchr() to find the extension in a filename.

Sigill:
Use strrchr() to locate the last dot in a filename instead of a homemade loop.


ggreer:
Thanks. I forgot about `strrchr()`.

PR #617, opened 2015-02-24
Checks if 'have' is declared before using it in bash completion script.

RedX2501:
This makes it possible to source this file in your .bashrc without getting an error message.

### DIFF
--- a/ag.bashcomp.sh
+++ b/ag.bashcomp.sh
@@ -1,115 +1,125 @@
-have ag &&
-_ag() {
-  local lngopt shtopt split=false
-  local cur prev
+# if you source this manually have will never exist as it is
+# undefed by bash_completion once it's done
+has_have=$(type -t have)
+has_ag=0
+if [[ $has_have == 'function' ]]; then
+  have ag
+  has_ag=$?
+fi
 
-  COMPREPLY=()
-  cur=$(_get_cword "=")
-  prev="${COMP_WORDS[COMP_CWORD-1]}"
+if (( $has_ag == 1 )); then
+  _ag() {
+    local lngopt shtopt split=false
+    local cur prev
 
-  _expand || return 0
+    COMPREPLY=()
+    cur=$(_get_cword "=")
+    prev="${COMP_WORDS[COMP_CWORD-1]}"
 
-  lngopt='
-    --ackmate
-    --all-text
-    --all-types
-    --after
-    --before
-    --break
-    --nobreak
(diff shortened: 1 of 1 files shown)


mcphail:
The issue with "have" is that it only exists if you have an outdated version of bash_completion. The current autotools setup doesn't handle the bash completion script well. I looked at doing this better (see PR #273) but there wasn't much interest at the time.

RedX2501:
@mcphail I checked out the latest bash-completion script to check when `have` was introduced and it is seems that it is still available and used.

mcphail:
https://anonscm.debian.org/cgit/bash-completion/bash-completion.git/commit/?id=20c05b43b6e44be8ae3c5025a77eca62a48c5a62

RedX2501:
You and i are both right i think... 

    $git rev-parse HEAD
    80c2bb696437134d6d8a3fb9c9f455fec75f470a
    
    $git grep have -- bash_completion
    bash_completion:19:#   You should have received a copy of the GNU General Public License
    bash_completion:116:# This function checks whether we have a given program on the system.
    bash_completion:118:_have()
    bash_completion:125:# Backwards compatibility for compat completions that use have().
    bash_completion:127:#             loaded completions, and _have is suitable for runtime use.
    bash_completion:128:have()
    bash_completion:130:    unset -v have
    bash_completion:131:    _have $1 && have=yes
    bash_completion:505:# See also: Bash FAQ - E13) Why does filename completion misbehave if a colon
    bash_completion:605:# `=' should have been removed from COMP_WORDBREAKS when setting $cur for
    bash_completion:717:    # Remove all redirections so completions don't have to deal with them.
    bash_completion:1514:    # If we have known_hosts files to use
    bash_completion:1627:    # we have a CDPATH, so loop on its contents
    bash_completion:1715:        # If we have no completion for $cmd yet, see if we have for basename
    bash_completion:1980:# completion files that may not have been sourced yet.
    bash_completion:2010:unset -f have
    bash_completion:2011:unset have

mcphail:
Yes - your fix is great but the way ag installs the bash completion is still wrong. If we want it to be dynamically loaded by bash-completion (>= version 1.99) we have to load it elsewhere and rename the script to "ag" rather than just stripping (or nullifying) the "have".

RedX2501:
My intentions with this patch was never to make it installable but only to make it runnable by hand, which is my scenario.

So i'll leave this here as is.

PR #625, opened 2015-03-07
Added -x/--line-regexp (match whole line) option

okdana:
`ag` already has the `-w` option, and this seems a fairly obvious complement to it, so i'm worried that this was requested and rejected at some point in the past. However, i can't find anything in the list of issues to confirm that so... here we go.

Just in case it's not super clear, `-x` is an option described in the POSIX specification for `grep` which causes the pattern to match on whole lines only. In other words, `grep -x 'foo'` has the same effect as `grep '^foo$'`.

This is an option that i miss a lot when using `ag`, so i thought i'd try to add it. Sorry if it's shit, this is only like the second time i've really done anything in C. I also added some cram tests for both this option and the existing `-w`.

```
Added -x/--line-regexp (match whole line) option
- Added -x functionality (like `grep`) for literal and non-literal matching
- Options -w and -x are mutually exclusive
- Added cram tests for -w and -x options
```


cup:
https://github.com/ggreer/the_silver_searcher/issues/626

PR #628, opened 2015-03-07
Treat literal '.' literally (fixes #627)

### LINKED ISSUE #627: Bug: Literal pattern '.' is not treated literally
Looks like there is a little optimisation for this pattern which causes searches for literal `.` to not be treated literally, even if `-Q` or `-F` is explicitly supplied.

Actual behaviour (in master):
```
% ./ag -F '.' <<< $'foo\nfoo.bar'
foo
foo.bar
```

Expected behaviour:
```
% ./ag -F '.' <<< $'foo\nfoo.bar'
foo.bar
```

okdana:
Fixes #627: *Bug: Literal pattern '.' is not treated literally*

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -36,7 +36,7 @@ void search_buf(const char *buf, const size_t buf_len,
         matches_spare = 0;
     }
 
-    if (opts.query_len == 1 && opts.query[0] == '.') {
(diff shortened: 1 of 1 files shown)


ggreer:
Nice find. Thanks for fixing it.

Issue #629, opened 2015-03-07
Feature request: make --version print compiled-time options


okdana:
Many utilities with functionality controlled by compile-time options print the options out when called with `--version`. Examples include `perl`, `rsync`, `vim`, &c. It might be useful if `ag` could do the same.

PR #630, opened 2015-03-07
Make --version show status of features controlled by compile-time options

okdana:
Addresses #629: *Feature request: make --version print compile-time options*

I am not very good at deciphering Autoconf/Automake stuff, so i wasn't sure which options should be printed, but i picked the three that seem the most relevant: PCRE JIT, lzma, and zlib. The result looks like this (i based the +/- thing on the way vim does it):

```
% ./ag --version
ag version 0.29.1

Features:
  +jit -lzma +zlib
```

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -106,7 +106,23 @@ For a list of supported file types run:\n\
 }
 
 void print_version(void) {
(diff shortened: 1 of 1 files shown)


ggreer:
Good idea. I might tweak the output a little after merging.

PR #632, opened 2015-03-08
windows: add native colors support

avih:
It might need some cleanups or refactoring, let me know what you think.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,12 @@
 ACLOCAL_AMFLAGS = ${ACLOCAL_FLAGS}
 
 bin_PROGRAMS = ag
-ag_SOURCES = src/ignore.c src/ignore.h src/log.c src/log.h src/options.c src/options.h src/print.c src/print.h src/scandir.c src/scandir.h src/search.c src/search.h src/lang.c src/lang.h src/util.c src/util.h src/decompress.c src/decompress.h src/uthash.h src/main.c
+ag_SOURCES = src/ignore.c src/ignore.h src/log.c src/log.h src/options.c src/win32/win_colors.c src/options.h src/print.c src/print.h src/scandir.c src/scandir.h src/search.c src/search.h src/lang.c src/lang.h src/util.c src/util.h src/decompress.c src/decompress.h src/uthash.h src/main.c
+
+# When cross-compiling with MXE, Makefile.w32 is not used but ag builds fine,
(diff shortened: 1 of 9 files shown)


avih:
`Makefile.w32` is missing a `\` after `src/util.c`. I added it but github might be having issues and the push doesn't work right now but I'll push it when I can. I don't have a platform to test `Makefile.w32` so I missed it.

ggreer:
Does this do the same thing as #333?

avih:
It solves the same problem eventually using the same API, but with slightly different approaches.

PR  #333 provides an alternative generic `fprintf` function which parses the ansi codes in whatever goes through it and tries to apply whatever it can to the screen, while my approach is more limited where I've split the existing `fprintf` calls into strings and colors, where the color parts may go through the ansi-to-windows interpreter.

I haven't followed all the other code, but I also tried to play nicely with the options, default options and fallbacks (on case of pager/pipe).

avih:
I actually have another patch to both simplify `set_color` and also make the colors handling more robust (can handle few more color codes, and overall more correct handling), but I'd appreciate feedback on the current approach first, and whether or not you wish to take it forward.

avih:
Hmm.. you merged PR #333 after I submitted this PR...

I just tried it and noticed these issues:
- It doesn't build with MXE using the default `Makefile.am` since `print_w32.c` is missing. I included it temporarily and it now builds.
- It doesn't highlight the matching pattern by default or with other values which I tried to use for `--color-match`. It does highlight the path and line though, and it is affected by `--color-line` and `--color-path`.
- It doesn't seem to distinguish between normal and bright/bold colors (seem to always use bright/bold), doesn't seem to "understand" blink or underline, and possibly few more.

All of those were handled correctly at my patch. Since you already landed it though, I'll try to fix the current code.

avih:
@mattn could you please comment on my findings above? are they only apparent on my build? or does your build exhibits them too?

mattn:
print_w32.c should work on win32 native not cygwin/msys2. cygwin/msys2 can handle ansi escape sequence it-self.
Currently, I build ag with `mingw32-make -f Makefile.w32`. But you seems to want autotools. I'll look into it in next week.

avih:
Updated the branch of this PR with some more pending updates I had, but closing this PR since #333 was merged.

I do have some fixes for it and I'll PR them.

PR #645, opened 2015-03-28
windows colors improvements

avih:
- can build also without `Makefile.w32`
- fixed pager/pipe colors - fallback to ansi colors for pager/pipe or CLI flag.
- doesn't show bold colors for everything, more compatible implementation.
- can show color also for matches (was only showing for path and line number).
- better includes, removed unused variables and static, initialize others.
- fixed undefined behavior if the print length is more than `BUFSIZ`
- some more stuff - details at the commit messages.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -14,6 +14,7 @@
 #include "lang.h"
 #include "log.h"
 #include "util.h"
+#include "print.h"
 
 const char *color_line_number = "\033[1;33m"; /* bold yellow */
 const char *color_match = "\033[30;43m";      /* black with yellow background */
@@ -42,6 +43,8 @@ Output Options:\n\
      --color-line-number  Color codes for line numbers (Default: 1;33)\n\
      --color-match        Color codes for result match numbers (Default: 30;43)\n\
      --color-path         Color codes for path names (Default: 1;32)\n\
+     --color-win-ansi     Use ansi colors on Windows even where we can use native\n\
+                          (pager/pipe colors are ansi regardless) (Default: off)\n\
      --column             Print column numbers in results\n\
      --[no]filename       Print file names (Enabled unless searching a single file)\n\
   -H --[no]heading        Print file names before each file's matches\n\
@@ -129,6 +132,7 @@ void init_options(void) {
     memset(&opts, 0, sizeof(opts));
     opts.casing = CASE_DEFAULT;
     opts.color = TRUE;
+    opts.color_win_ansi = FALSE;
     opts.max_matches_per_file = 0;
     opts.max_search_depth = DEFAULT_MAX_SEARCH_DEPTH;
     opts.path_sep = '\n';
@@ -215,6 +219,7 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
         { "color-line-number", required_argument, NULL, 0 },
         { "color-match", required_argument, NULL, 0 },
         { "color-path", required_argument, NULL, 0 },
+        { "color-win-ansi", no_argument, &opts.color_win_ansi, TRUE },
         { "column", no_argument, &opts.column, 1 },
         { "context", optional_argument, NULL, 'C' },
         { "count", no_argument, NULL, 'c' },
@@ -712,4 +717,8 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
     }
     (*paths)[i] = NULL;
     (*base_paths)[i] = NULL;
+
+#ifdef _WIN32
+     windows_use_ansi(opts.color_win_ansi);
+#endif
 }
--- a/src/options.c
+++ b/src/options.c
@@ -15,7 +15,7 @@
 #include "log.h"
 #include "util.h"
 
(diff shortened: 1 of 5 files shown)


avih:
@mattn, could you please verify that these changes don't break something I'm unaware of? Thanks.

ggreer:
Thanks for making this PR. Since I don't use Windows, I'd love to hear other Windows users chime in on this. Also, I'm finally adding clang-format tests to ag, so you'll have to rebase to get this mergeable again.

avih:
Not sure I understand the rebase comment, these patches are on top of the current tip. Or did you mean you *will* add clang-format tests, and after that these patches will have to be rebased?

As for the other changes, unused variables and no-op static declaration, assuming that a buffer is null terminated when it might not be, interpreting all color codes as bold (making it impossible to use non bold colors), not being able to use `--color-match` since it resets the color after each `fprintf` but `ag` prints the color in one statement and the match in another, the fact that colors are applied to the screen even when the output goes to another program (pager/pipe) - all those are just incorrect.

Do you expect other windows users to apply these patches, compile `ag` for windows (which is notoriously hard - the last publicly available version for windows is more than a year old and unmaintained), and then report back if these changes are worthwhile?

Despite the big number of patches, the changes to the code are few. It just fixes what was incorrect, and adds the `--color-win-ansi` option for those who want to bail out of using native windows colors if they use a terminal which support ansi colors.

avih:
> attr_initialized should take TRUE/FALSE. So should not be WORD.

Thanks. Done.

avih:
> so you'll have to rebase to get this mergeable again

Done.

mattn:
Where the `color_win_ansi` is used?

mattn:
Ah, I found it.

https://github.com/avih/the_silver_searcher/commit/37c49f0779724ca2ecb8c107f5c48efd4ec3ef5e#diff-08c125cc4d61b324db978312c589f3a4R39

avih:
Hmm.. it seems to me like the Travis CI failure is is unrelated to the patch. But if someone thinks otherwise, let me know.

src/options.c
```diff
@@ -14,8 +14,9 @@
 #include "lang.h"
 #include "log.h"
 #include "util.h"
+#include "print.h"
 
-const char *color_line_number = "\033[1;33m"; /* yellow with black background */
+const char *color_line_number = "\033[1;33m"; /* bold yellow */
```
ggreer: The color didn't change but the comment did?

src/options.c
```diff
@@ -42,6 +43,8 @@ Output Options:\n\
      --color-line-number  Color codes for line numbers (Default: 1;33)\n\
      --color-match        Color codes for result match numbers (Default: 30;43)\n\
      --color-path         Color codes for path names (Default: 1;32)\n\
+     --color-win-ansi     Use ansi colors on Windows even where we can use native\n\
```
ggreer: I think this belongs in the manpage, not the usage output. I try to keep usage as small as possible. It's no fun running `blah --help` and getting, "It was the best of times, it was the worst of times..."

src/options.c
avih: Yes I just fixed the comment. It's bold yellow, the same as `1;32` two lines below is bold green. Yellow on black background is `0;33`.

ggreer:
The Travis breakage was my fault. I've fixed it in master. I made a couple comments, but they're not show-stoppers. If you could rebase and move `--color-win-ansi` into the manpage, that'd be sweet. If that doesn't happen in a couple of days, I'll merge this and make the changes myself.

src/options.c
ggreer: Ah, got it. That's probably been wrong for years.

src/options.c
avih: Yeah, I should have added a man entry indeed. Will do. So you want it removed from the short help description?

src/options.c
avih: I'd have never have noticed it myself, but I happened to spent way too much time recently studying the ansi color escape sequences and how they behave on different systems, wrote an implementation from scratch, and then fixed another one ;)

src/options.c
ggreer: Yes. For >90% of ag users, this option doesn't matter. The needs of the many outweigh the needs of the few, or the windows user. Or something like that.

avih: Yeah, makes sense. That being said, it's unlikely that windows users will have a working `man` system, so they're likely to use the inline help almost exclusively. Would you be OK with something like this?
```
 ...
     --color-match        Color codes for result match numbers (Default: 30;43)\n\
     --color-path         Color codes for path names (Default: 1;32)\n\
"
#ifdef _WIN32
"\
     --color-use-ansi     Use ansi colors on even where we can use native (Windows)\n\
                          (pager/pipe colors are ansi regardless) (Default: off)\n\
"
#endif
"\
     --column             Print column numbers in results\n\
     --[no]filename       Print file names (Enabled unless searching a single file)\n\
 ...
```

I tested it and it works as expected. Though I could also end the previous `printf`, use another for the windows stuff, and then another for the rest.

PR #647, opened 2015-03-30
Add the Velocity file type

jonathanboom:
This adds the `--velocity` file type flag. I'm working in a codebase that makes heavy use of the [Velocity templating language](https://velocity.apache.org), and it would be very useful to search just the `.vm` files.


ggreer:
You forgot to update the test to notice the new output. I'll fix it, but please look at Travis CI status in the future.

jonathanboom:
Ah, sorry about that. Thanks!

PR #649, opened 2015-04-01
Added support for the Clojure 1.7 common file format

Leonidas-from-XIV:
This file extension is used for code common to all Clojure implementations, and will be supported starting with the soon to be released Clojure 1.7.


ggreer:
Sweet.

PR #650, opened 2015-04-06
Print WARN messages to stderr instead of stdout

apelisse:
When running ag on some path, it spits a lot of:

    WARN: Skipping ./long/path. Use the --depth option to search deeper.

I'm use to quiet grep(1) by appending `2>/dev/null`, but it just doesn't
work here because they're printed on standard output.

Move the warning to stderr so that it can be easily silented if
necessary.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -53,6 +53,7 @@ void vplog(const unsigned int level, const char *fmt, va_list args) {
             fprintf(stream, "MSG: ");
             break;
         case LOG_LEVEL_WARN:
+            stream = stderr;
             fprintf(stream, "WARN: ");
             break;
         case LOG_LEVEL_ERR:


ggreer:
I agree that the current behavior is undesirable, but I think it would make more sense to change the long path warning to an error. Here's how I think of the log levels:

1. DEBUG: Verbose debugging stuff.

2. LOG: Normal stuff.

3. WARN: Bad, but not so bad that it needs to go to stderr.

4. ERR: Pretty bad.

While `log_warn` isn't used anywhere else in the codebase, it could be handy in the future. Better to allow for different behavior than to make it `log_err` in all but name.

apelisse:
I'm usually not a big fan of logs, also for this reason.

I think I would move everything to stderr.
In my opinion stdout should be used for the actual output of the application, stderr for everything "meta". As an example, I might want to parse the output of `grep`/`ag` for whatever purpose, and I expect "messages" to not interfere with my file-descriptor. Let's say I do: `ag something | wc -l`. That'd be kind of awkward.

adamjernst:
I agree that WARN should go to stdout. Otherwise e.g. piping output into xargs breaks!

apelisse:
Wait, did you mean stdout or stderr? Having errors (or even warnings for that matter) spitted in stdout just makes it impossible to script properly

adamjernst:
Oh right I meant stderr!

PR #614, opened 2015-02-23
Properly handle gitignores.

RedX2501:
I kindly ask someone to test this with valgrind as i developed this patch under cygwin.
- Loads parent .gitignore if in git project
- gitignore from the current path are also loaded even if not in git
  project
- Respects gitignore priorities, more local gitignore patterns to
  searched file override further away patterns
- In-file pattern ordering is respected
- Include patterns work
- Added many tests to check for correct gitignore handling
- Not terribly slow (tested with linux-master c517d838eb7d07bbe9507871fab3931deccff539)

With SSD, bitlocker encrypted partition

```
fixed gitignores:
-----------------

$ time ag source > /dev/null

real    0m15.636s
user    0m7.627s
sys     0m20.202s

broken gitignores (master 1ed4631d7745134a29b5e1b5097dd7dba37c97c2):
---------------------------

$ time ag source > /dev/null

real    0m14.425s
user    0m7.550s
sys     0m19.279s
```

This uses wildmatch from git. According to  #356 this might be problematic due to copyright issues.

PR #654, opened 2015-04-10
Add tests for funky gitignore negations

RasmusWL:
Git has some funky features of [gitignore](https://git-scm.herokuapp.com/docs/gitignore)

```
dir/*
!dir/*/
```

The above gitignore means to ignore everything in `dir` except subdirs. However, if your parentdir is already ignored, no negation will change that fact.

This is a bit of a problem, as `fnmatch("dir/", "dir/*")` says there is a match, meaning `ag` will completely ignore that folder.

Handling gitignore negations as git does would really be awesome. I tried to implement this myself by adding a `char **noignore_regexes` to the `ignores` structure, and avoiding a match if one of these where valid, but that turned out to be harder than I thought. 

If your license is compatible, maybe you could just use the code from [`git-ls-files`](https://git-scm.herokuapp.com/docs/git-ls-files) or even libgit2 to get the files git cares about?

---

ps. Why are the patterns that need `fnmatch`  called `regexes`? I was a bit confused by this, as I thought the array would actually hold a regular expression.

### DIFF
--- a/tests/ignore_subcontent_except_onefile.t
+++ b/tests/ignore_subcontent_except_onefile.t
@@ -1,0 +1,28 @@
+Setup:
+
+# This is a test for gitignores, where you can ignore all subcontent
+# of a dir, but add exception to that.
+
+$ . $TESTDIR/setup.sh
+$ mkdir -p ./a/b
+$ mkdir -p ./a/c
+$ echo 'whatever1' > ./a/foo.txt
+$ echo 'whatever2' > ./a/important.txt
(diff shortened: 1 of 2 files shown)


RasmusWL:
Ahh, I see this was already mentioned in #614, and that you are not able to merge the code from git because of licensing.

However, please do consider adding the test cases supplied here and in #614, so it is obvious which test cases needs to be handled. Maybe even add a disclaimer that gitignores are not fully supported?

PR #660, opened 2015-04-21
Fix the test for `clang-format`

bdesham:
This consists of two changes:

1. The syntax in `configure.ac` was incorrect; the effect was that the build script always thought that the `clang-format` command was available, even if `$CLANG_FORMAT` was equal to `no`.
2. `format.sh` did not notice if `$CLANG_FORMAT` was not an actual command; in this case it would actually report that the code was formatted correctly and exit with a 0 status code. (This happened because the result of `clang-format` was being piped through `grep` and so the failure code from the invalid `clang-format` command was silently ignored.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -67,7 +67,7 @@ AC_CHECK_PROGS(
     [clang-format-3.7 clang-format-3.6 clang-format],
     [no]
 )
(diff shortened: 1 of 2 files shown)


ggreer:
Thanks so much for finding and fixing these bugs.

Issue #397, opened 2014-05-06
bashcomp: have ag?


ubermuda:
Hi there,

a few days ago, after upgrading all my brew installed packages – including ag – I started noticing this line when I open a new bash login shell:

```
-bash: have: command not found
```

After a bit of investigation, it appears it is caused by ag's `ag.bashcomp.sh` very first line:

```
have ag &&
```

I guess it is a command that tells whether a certain executable file is available in the current `$PATH`, unfortunately, my system doesn't have the `have` command (how ironic), and I can't find which package is supposed to have it. Couldn't find any package in homebrew (maybe I don't know how to search, that is a possibility). Damn, I can't even find anything on google.

Does anyone have any clue how I can fix that?

Issue #208, opened 2013-05-04
ag.bashcomp.sh issues with first line


danielb2:
`have ag &&`  does not work.  -> `bash: have: command not found`

removing first line makes the bash completion work fine.

PR #663, opened 2015-04-30
Bash-completion: add missing options, remove `have`

gentoo90:
`have` was mentioned in #208, #397, #617. It's only needed when `bashcomp` file is installed separately from the command to check if this command is present in a system.


ggreer:
I use zsh so I don't know much about this issue, but I'll trust your PR. Thanks :)

Issue #653, opened 2015-04-08
Improved `--stats` output 


Zearin:
I currently just used `--stats`, and good the following output:

```
702 matches
2668 files searched
38278054 bytes searched
1.211867 seconds
```

That’s pretty good.

But I think it could be better if it were like this:

```
702 matches
<NUMBER> files contained matches
2668 files searched
38278054 bytes searched
1.211867 seconds
```

For now, I can just use `--files-with-matches` and pipe the output to `wc -l` to get the number of files containing matches, but that’s a pretty clunky workaround.  

And besides—this is **the silver searcher**!  Badass statistics are what it’s all about. :-)

openaphid:
+1 for number of files in stats

ggreer:
Added to master. Thanks for the idea and thank you @epmatsw for implementing it.

PR #664, opened 2015-04-30
Add number of files matched to --stats output

epmatsw:
As requested in issue #653, this adds the number of matched files to the output for --stats.


ggreer:
Thanks!

Issue #168, opened 2013-03-05
! in .gitignore isn't taken into account.


rubikitch:
$ cd /tmp; mkdir agt; cd agt
$ echo a > f1 > f2
$ echo f1 > .gitignore
$ ag --nocolor --nogroup a 
f2:1:a
$ (echo '*'; echo !f2) > .gitignore
$ ag --nocolor --nogroup a 
# Should be 'f2:1:a' here!!

$ rm -rf /tmp/agt

sloria:
To add to @austinpray's comment, any file that is gitignored using star-globbing is not ignored by ag:

```
# ag will not ignore files in build/ =(
build/*

# but this works
build/
```

Would love to see this bug fixed.

jasonkarns:
As commented on a [couple](https://github.com/ggreer/the_silver_searcher/issues/594#issuecomment-85543490) of [related](https://github.com/ggreer/the_silver_searcher/issues/530#issuecomment-85544199) issues, this could be solved if ag got it's git ignore list from `git status --ignored`

blueyed:
@jasonkarns 
Using git's interface is a good idea, but I think calling `git` would be bad in regards to performance, especially on Windows probably.
Instead, it could / should probably be retrieved using libgit2 or something similar.

JacksonGariety:
Any status on this? Workaround? It's annoying.

luochen1990:
I hit this issue when solving [another problem](https://stackoverflow.com/questions/19023550/how-do-i-add-files-without-dots-in-them-all-extension-less-files-to-the-gitign/19023985#comment98881583_19023985) which requires an exclude-and-re-include solution.

I'm wondering is there any progress here since several years has passed :)

jomi-se:
https://github.com/ggreer/the_silver_searcher/pull/1101 was merged a while ago but it doesn't seem to have solved this issue 😞  (or maybe it was reverted at some point 🤔 )

YorkZ:
I just compiled commit 965f71d, and tested with the following .gitignore file:

```
*.d
!*.d/
```
But `ag` still ignored my directory `./rootfs/packages.d`.

Integralist:
Oh dear. An almost 8 year old issue is never good 🙂 

Seems for me the solution is to use `-U --skip-vcs-ignores`, hopefully that helps someone else who stumbles across this.

YorkZ:
[...]

PR #665, opened 2015-05-02
POC fix for gitignore

flaviut:
This patch isn't in a usable state, but it functions as a proof of concept fix for the gitignore problems. It is absolutely not in a mergable state, reviewing the code is pointless at this point in time.

It works by asking git what files should be ignored, using the `git status --ignored --porcelain` command. I can confirm it fixes at least #168.

Possible objections:

- the dependency on git might not be wanted, although there isn't much point to searching a git repo without git installed.
- potential performance issues

-------

If this is an acceptable way of solving these issues, I'd be happy to properly recreate this PR.

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -186,6 +186,27 @@ void load_ignore_patterns(ignores *ig, const char *path) {
     fclose(fp);
 }
 
+void load_ignored_git(ignores *ig, FILE *stdout_fp) {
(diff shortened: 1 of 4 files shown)


flaviut:
Looks like I wasn't the first one to come up with this idea, @jasonkarns mentioned it in the comments on #168.

I'll look into using libgit2.

PR #672, opened 2015-05-21
Support patterns beginning with a hyphen (grep -e analogue)

bishopb:
I have found myself searching for portions of object chains, like "->thread->get", but couldn't do that because the dash in the pattern was interpreted as an option.  In my home grown solution pre-ag, I used `grep -e` always, so I thought ag could get this little enhancement.  Unlike grep -e though, multiple patterns aren't supported.

Examples of new support:
- `ag -e '->thread'`
- `ag -e '->thread->'`
- `ag -e '->thread.*->'`


bishopb:
Build fails because of:

> make: *** No rule to make target `doc/ag.1', needed by `all-am'.  Stop.

Should PR include regenerated `doc/ag.1`?

aaronbauman:
+1 for this

bishopb:
@aaronbauman While I'd like to have this for argument compatibility with `grep`, I've pretty much just been using the GNU bare double dash workaround: `ag -- '->foo('`.

aaronbauman:
that works - thanks for the tip

PR #676, opened 2015-05-27
Fix build warning

### LINKED ISSUE #662: Compile Warning
Building from c81622c5c5313c05eab2da3b5eca6c118b74369e, I'm getting a warning during compilation.

CC       src/decompress.o
src/decompress.c:51:20: warning: assigning to 'Bytef *' (aka 'unsigned char *')
      from 'const void *' discards qualifiers
      [-Wincompatible-pointer-types-discards-qualifiers]
    stream.next_in = buf;
                   ^ ~~~
1 warning generated.

I'm also getting a "clang: warning: argument unused during compilation: '-pthread'", but I don't think that that's related.

My gcc --version:
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 6.1.0 (clang-602.0.49) (based on LLVM 3.6.0svn)
Target: x86_64-apple-darwin14.4.0
Thread model: posix

thenewvu:
Just a minor change to fix #662 
That issue happens on my side too (Ubuntu 14.04 x64).

### DIFF
--- a/src/decompress.c
+++ b/src/decompress.c
@@ -49,7 +49,7 @@ static void *decompress_zlib(const void *buf, const int buf_len,
 
     stream.avail_in = buf_len;
     /* Explicitly cast away the const-ness of buf */
-    stream.next_in = (Bytef *)buf;
+    stream.next_in = (const Bytef *)buf;
 
     pagesize = getpagesize();
     result_size = ((buf_len + pagesize - 1) & ~(pagesize - 1));


epmatsw:
Probably either should remove the comment on the line above (`/* Explicitly cast away the const-ness of buf */`), or figure out why that cast was added in the first place.

blackb1rd:
According to http://www.zlib.net/manual.html 

```c
z_const Bytef *next_in;     /* next input byte */
```

This should be cast as const Bytef * instead of Bytef *.

eoranged:
It seems that comment line above is not needed anymore.

PR #679, opened 2015-05-28
Add -e/--pattern option

rzlourenco:
grep provides an option to use an explicit pattern. This is useful since we may need to search for a pattern that starts with `-` or `--`.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -79,6 +79,7 @@ Search Options:\n\
                           or patterns from ignore files)\n\
   -D --debug              Ridiculous debugging (probably not useful)\n\
      --depth NUM          Search up to NUM directories deep (Default: 25)\n\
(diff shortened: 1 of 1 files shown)


RomanHargrave:
As far as my experience goes, this can be done already:

`ag --fish -- '--flag' scripts/` should work, as it would with GNU grep `make install | grep -v -- '-- Up-to-date'`.

bishopb:
See also the earlier PR https://github.com/ggreer/the_silver_searcher/pull/672 that already implemented this.  That said, the `--` workaround by @RomanHargrave makes sense.

wsdjeg:
I still prefer use an option instead of seperator. I just check ripgrep, it support `-e` option, after `-e` option I can use any string.

orgads:
:+1:

Issue #678, opened 2015-05-28
Files with extensions with multiple dots are not recognized in .gitignore


xbenes:
Having the following `.gitignore`,
```
*.backup.txt
```
I would expect not searching in such files.

Reproducible example: having above mentioned `.gitignore` and a file named for example `info.backup.txt`, ag searches its content.

Note: tried with current master, it seemed to work fine with v0.19.2

epmatsw:
`git bisect` pinpoints this as introduced in c6a4f20a34d44feaa95eb1aa582155573944d106.

AshleyTharpTL:
Any ideas for temporary workarounds?

slamorte:
Bug also exists for whitelisting.
```
# .gitignore
*.gz
*.sql

!foo.gz  #file is committed
!foo.sql   #file is committed
!foo.sql.gz   #foo.sql.gz remains ignored
!foo.*   #foo.gz and foo.sql whitelisted, foo.sql.gz remains ignored
```

axemaster:
also seeing this issue.

PR #680, opened 2015-05-28
Fix Extension Matching

epmatsw:
Extension ignoring is kind of broken when there are more than one .'s in the ignore pattern. We find the extension by using a `strrchr` to search for the last `.` and then try to match the characters after that. So, for example, *.backup.txt ends up with an extension of "txt", which fails to match any of our patterns. This means that the file is not ignored when it should be (issue #678).

This is a simple fix of just searching forward in the string rather than backward. It's possible that we don't want to include *.ext.otherext patterns in the extensions array, in which case we need a different fix that is probably over my head. I've also included tests for this fix.

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -381,7 +381,7 @@ int filename_filter(const char *path, const struct dirent *dir, void *baton) {
     }
     log_debug("path_start %s filename %s", path_start, filename);
 
-    const char *extension = strrchr(filename, '.');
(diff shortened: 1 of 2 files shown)


ggreer:
Thank you for the test case. It's a shame some file "extensions" have multiple dots, as this does impact performance a little. Oh well. Enough users seem to be running into this issue that the minor performance loss is worth the correct behavior.

Issue #677, opened 2015-05-27
Huge column numbers when matching newline


Szunti:
With the file:
```
a
b
```

```
$ ag --column 'a\s' /tmp/test.ag
1:1:a
2:18446744073709551615:b
```
is the result.
#459 may be related.

ggreer:
This should be fixed in master.

PR #681, opened 2015-05-28
Avoid underflow in column count on multiline matches

epmatsw:
When --column is enabled, the starting column of a match is stored as an unsigned long. When the match spans multiple lines, the calculated starting column is actually negative, which results in an underflow and a really large column number like #677. This fix changes the behavior to print 0 as the column number rather than calculating and printing an incorrect unsigned long.


ggreer:
I fixed this and added test cases. I apologize for not merging this. I thought of a simpler way to implement the same thing.

Thank you for helping though. I wouldn't have fixed this if not for your PR.

epmatsw:
No worries! Like you said, your fix was better :)

PR #689, opened 2015-06-12
make log output be MT-safe

mogutan88:
Hi, While I am using ag for my projects I see a strange behavior on ignore file and -t option, and started investing.
During that investigation I used `ag -D` and found that the debug output is not MT-safe(a log from a worker thread may be put in the middle of another worker log output), thus I saw the debug output like below,

```
DEBUG: DEBUG: DEBUG: DEBUG: DEBUG: DEBUG: DEBUG: No match in ./t/internal/ganesha/stamp/create.tDEBUG: No match in ./t/internal/ganesha/stamp/delete.tNo match in ./t/internal/ganesha/stamp/update.tNo match in ./t/internal/ganesha/stamp/fetch.tNo match in ./t/internal/ganesha/stampcategory/create.tNo match in ./t/internal/ganesha/stampcategory/delete.tNo match in ./t/internal/ganesha/stampcategory/fetch.t
Skipping ignore file ./t/restful/v2/.git/info/exclude: not readable
```
 So I made a quick patch to fix this problem.

Confirmed this worked on my OS X (10.9 and 10.10) and this passed crum test too.
Please review!

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -43,26 +43,27 @@ void vplog(const unsigned int level, const char *fmt, va_list args) {
         return;
     }
 
+    char *heading;
     FILE *stream = out_fd;
 
     switch (level) {
         case LOG_LEVEL_DEBUG:
-            fprintf(stream, "DEBUG: ");
+            heading = "DEBUG: ";
             break;
         case LOG_LEVEL_MSG:
-            fprintf(stream, "MSG: ");
+            heading = "MSG: ";
             break;
         case LOG_LEVEL_WARN:
-            fprintf(stream, "WARN: ");
+            heading = "WARN: ";
             break;
         case LOG_LEVEL_ERR:
             stream = stderr;
-            fprintf(stream, "ERR: ");
+            heading = "ERR: ";
             break;
     }
-
-    vfprintf(stream, fmt, args);
-    fprintf(stream, "\n");
+    char full_fmt[strlen(heading) + strlen(fmt) + 2];
+    sprintf(full_fmt, "%s%s\n", heading, fmt);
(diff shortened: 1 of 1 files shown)


src/log.c
```diff
@@ -43,26 +43,27 @@ void vplog(const unsigned int level, const char *fmt, va_list args) {
...
         case LOG_LEVEL_ERR:
             stream = stderr;
-            fprintf(stream, "ERR: ");
+            heading = "ERR: ";
             break;
     }
-
-    vfprintf(stream, fmt, args);
-    fprintf(stream, "\n");
+    char full_fmt[strlen(heading) + strlen(fmt) + 2];
```
ggreer: Variable-length arrays aren't supported in C89.

mogutan88:
@ggreer Replaced the variable-length array with malloc/free'd buffer.
Also I found that the previous fix did guarantee the atomicity only among vplog(s), but not for other fprintf(s) in print.c/search.c etc.
So introduced mutex lock into log.c as search.c has done so already.

ggreer:
Cool.

Though if you're going to lock a mutex, why not keep the simplified `fprintf()` and `vfprintf()` calls?

mogutan88:
@ggreer yeah, I actually wondered to revert the initial change(8df53ca) and simply use mutex lock when I tried to submit the 2nd commit (151e9c5).
I left sprintf (8df53ca) simply because it minimizes the locking time and the number of [v]fprintf() calls as well.

If you prefer not changing the original fprintf() and vfprintf() calls sequence, I will follow you so just let me know.

PR #691, opened 2015-06-16
README: Use the svg Floobits image

mc10:
Sadly, I could not find the Shields svg image.


ggreer:
I got a retina display last month, but I haven't looked at the ag readme until now. Those are some blurry images.

PR #700, opened 2015-06-23
Add support for .cjsx files under --coffee

epmatsw:
[Coffee-React](https://github.com/jsdf/coffee-react) allows CoffeeScript users to write JSX with more CS-like syntax. Since .jsx is supported under the --js flag, it seems reasonable that .cjsx should be supported under the --coffee flag.


ggreer:
Thanks!

Issue #189, opened 2013-04-21
--print-long-lines is documented but not implemented


dandavison:
I believe that `--print-long-lines` is not actually implemented, despite being documented in `ag --help` and `man ag`. There's a [TODO comment](https://github.com/ggreer/the_silver_searcher/blob/master/src/options.h#L47) in options.h indicating that it's not implemented. Should either implement or undocument.

I'd find it useful (was getting annoying hits from minified js files that lack the .min.js extension).

amfarrell:
I too would like to be able to either filter out really long lines or to have that be the default and to be able to include them again.

amfarrell:
One way to get around this is something like `ag '.csv' | awk '{ if (length($0) < 2028) print }'`, but it removes ag's nice formatting.

amfarrell:
One reason why this behavior would be useful is that often, one is using the silver searcher to look through a project that contains minified js alongside readable code. The user would prefer to omit the terminal-filling block of unreadable js from the results so they can focus on the readable code alongside it.

amfarrell:
I've submitted https://github.com/ggreer/the_silver_searcher/pull/707 to fix this.

AtomicNess123:
Would it be possible to specify how many characters before and after the match? This is ideal for, as mentioned above, prose text. Otherwise, one gets huge chunks of results which makes it troublesome to read. Thanks!

PR #706, opened 2015-06-30
Remove NOP option print-long-lines

amfarrell:
The options --print-long-lines do not actually do anything.
The silver searcher's default behavior is to print long lines.
While it would be really nice if by default it did not and this option
was required to turn that on , that has not been implemented. 
(see issue #189 )
Having this option is misleading to folks that are looking for a way
to have ag remove the long lines from the output.

### DIFF
--- a/ag.bashcomp.sh
+++ b/ag.bashcomp.sh
@@ -68,7 +68,6 @@ _ag() {
     --passthrough
     --passthru
     --path-to-agignore
-    --print-long-lines
     --print0
     --recurse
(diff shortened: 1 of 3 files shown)


amfarrell:
I would prefer this not be merged and that https://github.com/ggreer/the_silver_searcher/pull/707, which adds the missing functionality, be merged instead.

rainerborene:
+1

amfarrell:
This is now a duplicate of #910

PR #707, opened 2015-06-30
Skip long lines (fixes #189)

### LINKED ISSUE #189: --print-long-lines is documented but not implemented

amfarrell:
This actually implements the functionality to ignore lines longer than some number of characters (default 2000) and adds an option --long-line-length to change the that number. Tested on OS X Yosemite.


andrewthad:
If this actually works, I'm very excited.

amfarrell:
If you want to test this now, you should be able to 

    git clone https://github.com/amfarrell/the_silver_searcher/ -b ignore-long-lines
    cd ignore-long-lines

and then run through https://github.com/amfarrell/the_silver_searcher/tree/ignore-long-lines#building-master to install from source

epmatsw:
Tests aren't passing due to clang-format on my machine :( All functional tests pass though! Probably also should have some new tests in place to verify that this works as intended, but it looks awesome if it works! :+1:

amfarrell:
@epmatsw I've added a test, though I am note sure how to get it to run. `make test` doesn't do anything.

epmatsw:
@amfarrell That's interesting. No errors or anything? You'll need to have [cram](https://bitheap.org/cram/) installed, but other than that I think it should just work if you do `./build.sh && make test`

amfarrell:
ah, that was the problem. I had not installed cram.

amfarrell:
Is anyone else able to view the Travis CI results to see what failed? Building on my local machine works.

epmatsw:
Mine are failing locally, but still on clang-format. All functional tests (including the new one) are passing locally. Travis is failing on functional tests though, which is confusing.

andrewthad:
The [functional test it's failing on](https://travis-ci.org/ggreer/the_silver_searcher/jobs/69004907#L454) is the one that @amfarrell added. It doesn't seem like any of the other tests are failing on travis.

alanshutko:
That functional test is succeeding on my local El Cap box, too.  I wonder what the difference is between our machines and Travis?

pda:
I'm also keen on having this default option, mostly for the minified JS use-case, also some XML, JSON etc. I went looking for it in the man page, found it documented, read the source, saw it was a TODO, investigated implementing it, saw it needs to be deep within the print logic, then found this PR. :+1: 

My bike-shed input would be to reduce the default threshold to e.g. 1,000 chars; even that's 13 lines of text in ye olde 80-wide terminal.

My concern is that calling this new `get_line_length()` function for every matched line could be a performance regression, but I haven't tested that. It's probably duplicating some of the work done to actually print the line. Thoughts?

hchargois:
I hope this gets implemented :)
It would be nice if, by default, the `long-line-length` equals the (detected) width of the terminal, so that result lines fit nicely within the terminal, and never break to a new line.

pda:
> It would be nice if, by default, the long-line-length equals the (detected) width of the terminal

Sounds like you're looking for `--width` / `-W` from https://github.com/ggreer/the_silver_searcher/pull/720 which changes how matches are displayed.

Whereas my understanding of this `--print-long-lines` / `--long-line-length` option is that it completely excludes matches from very long lines (e.g. minified version of a JS file); I wouldn't want that defaulting to to the width of my terminal, else `ag` would constantly be failing to find matches in reasonable length lines.

In fact I think `-W` in #720 does a better job at what the never-released `--print-long-lines` / `--long-line-length` feature was aiming for; avoids sending many kb of text to the terminal when matches are found in long line files.

hchargois:
> Sounds like you're looking for --width / -W from #720 which changes how matches are displayed.

Yup, you are right, thanks! It works as I hoped, apart from the "auto-detect the term width" feature. Sorry for the noise.

> In fact I think -W in #720 does a better job at what the never-released --print-long-lines / --long-line-length feature was aiming for; avoids sending many kb of text to the terminal when matches are found in long line files.

I agree, I think people want to avoid things like minified JS flooding their terminal. That's why I was looking for this option. And I think it's best if you can see _concisely_ that those files match, I don't want them completely hidden.

amfarrell:
I was never able to get that test to fail locally. Given that the --width option was added in #720, I'm going to whistfully close this PR.

PR #257, opened 2013-09-07
Add support for ~/.agrc default config options

schlosna:
I can't live without my default options, so this adds support for ~/.agrc similar to .ackrc

* Load .agrc configuration file from user's home directory if available.
* CLI options take precedence over .agrc options

PR #701, opened 2015-06-25
Ag config file support implemented

decaff:
I took David Schlosnagle's code (@schlosna) and adapted it to follow many
of the suggestions discussed in issue #257.  Changes between David's code and
my adaptation:

1) Almost 100% of the code lives in 2 new source files.  src/options.c
   is not touched.
2) Consequently, merging this code with current forks is pretty easy.
3) Also due to 1) above, if you know C, you can hack the config file features
   to your heart's content.
4) Multiple config file locations are tested at startup (first found file
   wins, all others ignored). Here are the locations and order of access:

	1.   Path specified in $AGRC_PATH
	2.   $XDG_CONFIG_HOME/agrc
	3.   ~/.agrc
	4.   /usr/local/etc/agrc
	5.   /etc/agrc

5) Config file format:

	# this is a comment => inline comments not supported
	# blank lines okay as well

	--long-option[=value]
	-short-option[=value]
	--long-option[ value]
	-short-option[ value]

	# And special keywords (of which there is only one)
	list

Actual Example File (mine)
	# set up nice colors for a white terminal bkgnd...

	# bold blue
	--color-line-number=1;34
	# black with grey bkgnd
	--color-match=30;47
	# green
	--color-path=32

The features and paths are documented in the ag man page.

Issue #79, opened 2012-09-28
Environment variable for default parameters


nogweii:
I could create an alias, which is what I'll do for now, but it would be pretty awesome if you could pass parameters to ag via an environment variable. sort of like less's LESS or ack's ACK_OPTIONS. I would guess including them in the beginning of the string before passing argv to parse_options would make the most sense.

lgarron:
Anyone looking for the latest status: head straight to #257 (canonical bug number for this issue) and/or #709 (latest pull request).

stfl:
+1

PR #709, opened 2015-07-03
Add Configure File Support - Version 2

decaff:
Implementation is based on PR #257, authored by David Schlosnagle.

Ag reads OPTIONS from one of several configuration files.  Ag processes "config
file"  options  before  command-line options.  The config file format is very
simple:

    # this is a comment => inline comments not supported
    # blank lines okay as well

    --long-option[=value]
    -short-option[=value]
    --long-option[ value]
    -short-option[ value]

    # And special keywords (of which there is only one)
    list

        Actual Example Config File
        ==========================
        # set up nice colors for a white terminal bkgnd...

        # bold blue
        --color-line-number=1;34
        # black with grey bkgnd
        --color-match=30;47
        # green
        --color-path=32

If the keyword "list" is included  in a config file, ag dumps the  contents of
argv[]  on stdout.  This is useful for debugging config file errors that cause
ag to output a usage message and then silently exit.

Ag reads configuration information from a number of possible locations,
in the following order (first file found is used, all others ignored):

    1.   Path specified in $AGRC_PATH
    2.   $XDG_CONFIG_HOME/agrc
    3.   ~/.agrc
    4.   /usr/local/etc/agrc
    5.   /etc/agrc

### DIFF
--- a/doc/ag.1
+++ b/doc/ag.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "AG" "1" "February 2015" "" ""
+.TH "AG" "1" "July 2015" "" ""
 .
 .SH "NAME"
 \fBag\fR \- The Silver Searcher\. Like ack, but faster\.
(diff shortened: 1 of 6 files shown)


aisapatino:
@ggreer thoughts on this? or some other way of supporting default options? #574

907th:
+1 Waiting for it!

deivid-rodriguez:
@ggreer Are you considering this?

jondavidjohn:
Big fan of this idea!

modelm:
I'm also eagerly awaiting this feature. I use ag.vim and have to set my preferred options as both a bash alias and separately in my vim configuration at the moment.

vampolo:
@ggreer fell on this today. any chance it will be merged?

rosshadden:
:+1: This would be a lot better than aliasing `ag` to `ag --various --options`.

lgarron:
I would also love to see a defaults file. I have had trouble switching to `ag` instead of `grep` because I kept missing results – I just found out today I should always pass in the `--hidden` flag to help with that (and then use `.agignore` to exclude`.git` dirs).

For what it's worth, I got here via (!):

1. Google search for "[the_silver_searcher default arguments](https://www.google.com/search?q=the_silver_searcher+default+arguments
)" (second result) 
2. #79 ("Actually, I've changed my mind about this feature. It really seems like aliases are the way to go.")
3. #121
4. #257
5. #701

It would be nice if the file could be located either at `~/.agrc` *or* under `~/.config/`, as `git` and many other tools allow. But I guess `.agignore` doesn't support that either.

kugel-:
@ggreer please consider this. I'm coming from ack and truly miss this.

stfl:
+1

ZyX-I:
I would say that current PR has some flaws:

1. ~/.agrc should not be present present at all: XDG spec does not expect this.
2. Yet I would suggest to use ./.agrc, ../.argrc, ../../.agrc, etc; in many cases this would also pickup ~/.agrc.
3. Need a command-line option to ignore agrc files completely. Alternatively: command-line option to specify one agrc file which will be used in place of all defaults, AGRC_PATH is not needed much in this case.
3. What /usr/local/etc/agrc and /etc/agrc are doing there? Should be using `${XDG_CONFIG_DIRS:-/etc/xdg}`, `$xdg_config_dir/agrc` (XDG_CONFIG_DIRS is a list): XDG spec also cares about location of system-wide configuration.

doc/ag.1
```diff
@@ -243,5 +243,79 @@ Use the \fB\-t\fR option to search all text files; \fB\-a\fR to search all files
...
+.
+.SH "CONFIGURATION FILE LOCATIONS"
+Ag reads configuration information from a number of possible locations, in
+the following order (first file found is used, all others ignored):
+.RS
+.P
+.nf
+1.   \fIPath specified in $AGRC_PATH\fR
+.br
+2.   \fI$XDG_CONFIG_HOME/agrc\fR
```
jxpx777: I would strongly suggest this not be put directly into the `XDG_CONFIG_HOME`. Instead, I would suggest using a subdirectory of `XDG_CONFIG_HOME` the same way git and others do: `$XDG_CONFIG_HOME/ag/agrc`. It's been suggested in #1020 that the ignore lookup should also conform to XDG spec, and having these all in a single directory makes more sense than proliferating directly in the root of `$XDG_CONFIG_HOME`

PR #717, opened 2015-07-23
Added mako as filetype

amix:
Mako is one of the most popular Python template systems.


ggreer:
Next time, please update the output test too.

PR #720, opened 2015-07-30
Added an option to truncate long lines with -W NUM

shabbyrobe:
I'm doing a lot of agging over a codebase with single line packed JS files scattered through it. I do want to know if there were matches, and on which line, but some of those files with one massive line can play havoc with the output when I'm trying to navigate a couple of dozen matches. Things can get a bit tricky when most of the matches are nice 100 column lines, then one 15,000 column monster comes along! Lines that long don't exactly play nice with a terminal or a text editor (it crashes the current master of neovim using the Ag.vim plugin).

I've solved my problem by adding a `-W NUM` / `--width NUM` option which truncates matched lines to the length specified. -W was chosen as it didn't seem to clash with anything else, including in `grep` and `ack`. I thought I might clean it up, write some tests and see if you might find it a useful or worthy addition to ag. `make test` runs without errors, but I think my own test case might be missing a few scenarios, like the truncation when `print_path` is set to `PATH_PRINT_EACH_LINE`.

There are a couple of warts with my implementation - the newline handling on line 205 of print.c is problematic when you redirect the output to a file. It seems to clash with the "file without newline" detection. I didn't want to start digging too deeply into the newline handling logic without first starting a discussion. I figured it's also one of those things where someone who knew the codebase more intimately might be able to say "oh yeah, that's easy: you're doing that bit wrong, just move it over here and pow!", or even just "no, this whole thing isn't a good idea, you can already do that using *x*" ;)

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -103,6 +103,7 @@ Search Options:\n\
                           (.gitignore, .hgignore, .svnignore; still obey .agignore)\n\
   -v --invert-match\n\
   -w --word-regexp        Only match whole words\n\
+  -W --width NUM          Truncate match lines after NUM characters\n\
   -z --search-zip         Search contents of compressed (e.g., gzip) files\n\
 \n");
     printf("File Types:\n\
@@ -141,6 +142,7 @@ void init_options(void) {
     opts.color_win_ansi = FALSE;
     opts.max_matches_per_file = 0;
     opts.max_search_depth = DEFAULT_MAX_SEARCH_DEPTH;
+    opts.width = 0;
     opts.path_sep = '\n';
     opts.print_break = TRUE;
     opts.print_path = PATH_PRINT_DEFAULT;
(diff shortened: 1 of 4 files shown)


ggreer:
Thanks. I had to fix some conflicts and merge this manually.

shabbyrobe:
Thanks for merging! Sorry about the conflicts. Did you have any thoughts on my issues with the output redirection?

topher200:
On ag version 1.0.1 on macOS Sierra. My `man` page doesn't mention `--width` at all.

Not sure if this is an oversight or by design. I would have expected to see it in there - I was surprised to find this PR as the only information for this (very helpful!) option.

ayoon-toast:
Tuning in 5 months later - the `--print-long-lines option` is still in the man page, and no mention of `--width`

exebook:
As of April 2017 Ubuntu's default version of ag -- 0.31.0, has neither of options working.

PR #728, opened 2015-08-10
add support for BitBake source files

cardoe:
Add a --bitbake option which will search through the following files:

- .bb
- .bbappend
- .bbclass
- .inc
- .conf

### DIFF
--- a/src/lang.c
+++ b/src/lang.c
@@ -9,6 +9,7 @@ lang_spec_t langs[] = {
     { "ada", { "ada", "adb", "ads" } },
     { "asm", { "asm", "s" } },
     { "batch", { "bat", "cmd" } },
+    { "bitbake", { "bb", "bbappend", "bbclass", "inc" } },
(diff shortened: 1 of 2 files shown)


ggreer:
Won't there be a lot of false positives from matching .conf? I have lots of .conf files in my `~/code`, but none of them are related to BitBake.

cardoe:
On 9/11/15 1:45 PM, Geoff Greer wrote:
> Won't there be a lot of false positives from matching .conf? I have lots
> of .conf files in my |~/code|, but none of them are related to BitBake.
> 

I wouldn't be be opposed to dropping .conf from the patch.

-- 
Doug Goldstein

ggreer:
Ok. Feel free to change that and I'll merge it.

Sorry for the delay in responding to this PR. Work is a pretty big time-sink.

ggreer:
Thanks!

PR #729, opened 2015-08-11
Add filetypes for the bro framework

jzipfler:
Bro (www.bro.org) is a network security monitor with its own scripting language (.bro) and the ability to provide functionalities from C/C++ (.bif) mixed with the own language.

I only added these two filetypes to the lang.c and adjusted the related test.


ggreer:
Thanks!

Issue #699, opened 2015-06-22
Overall count


adius:
Is there no way to get an overall count of matches?
Something like: 

```sh
ag -C 0 --nogroup test | wc -l
```

This is an absolute must-have-feature in my opinion.

epmatsw:
Does --stats not work?
```
1613 matches
290 files contained matches
2944 files searched
12485981 bytes searched
3.294397 seconds
```

SnoringFrog:
Stats is the best way to get this that I've found. Piping ag into `ag match | tr -d " matches` is the simplest way I've come up with to get just the raw number if you need it in a script.

What would be nice is if including the `--nofilenames` option alongside `-cl` would produce this output, like how `ack -clh` or `ack -cl --no-filename` works.

SnoringFrog:
There is also now a `--stats-only` option that will only print the stats and not any of the matches themselves, which is even closer to what you were looking for.

PR #730, opened 2015-08-13
Combining -c and --nofilename prints just the number of matches. Similar to ack's -clh

SnoringFrog:
Addresses issue #699. 

With this, combining `-c` and `--nofilename` prints just the total number of matches. 

Note: if `--nofilename` is followed at any point by a header option while doing this, then the file names will still be printed, which is about the same as what would happen when you combine those two in that order without `-c`, but it might *feel* like a bug if you're thinking you'll only get the total number of matches.


SnoringFrog:
Whoops. Looks like this prints only the total matches anytime -c is being used, based on the tests (which I can't run locally because I can't get cram installed).

SnoringFrog:
Realized this functionality breaks some things you just fixed (installed via brew, so I don't think those changes were in what I originally got, and I never noticed them when I forked), closing this request.

PR #733, opened 2015-08-19
Add --stats-only option

SnoringFrog:
Allows you to print just the stats and nothing else. Good way to check things like total matches from a script, or manually without flooding the screen.


ggreer:
Thanks!

PR #735, opened 2015-08-23
Add faster substring search

acconsta:
This commit implements the fast substring search algorithm described by
Leonid Volnitsky.

The full algorithm is described at:
http://volnitsky.com/project/str_search/

Briefly, we insert each consecutive pair of characters in the needle
into a hash table. We then step through pairs of characters in the
haystack. If the pair is in the hash table (and probably also the needle)
we do a character-by-character comparison.

Performance is 30-40% faster than Boyer-Moore on 2GB of text:

> ./ag_hash --stats blahblahbla ~/Downloads/wiki/
> 0 matches
> 0 files contained matches
> 4 files searched
> 1956034523 bytes searched
> 0.800452 seconds
> 
> ./ag_master --stats blahblahbla ~/Downloads/wiki/
> 0 matches
> 0 files contained matches
> 4 files searched
> 1956034523 bytes searched
> 1.310972 seconds

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -109,6 +109,7 @@ int main(int argc, char **argv) {
         generate_alpha_skip(opts.query, opts.query_len, alpha_skip_lookup, opts.casing == CASE_SENSITIVE);
         find_skip_lookup = NULL;
         generate_find_skip(opts.query, opts.query_len, &find_skip_lookup, opts.casing == CASE_SENSITIVE);
+        generate_hash(opts.query, opts.query_len, h_table, opts.casing == CASE_SENSITIVE);
         if (opts.word_regexp) {
             init_wordchar_table();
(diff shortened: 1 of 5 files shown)


ggreer:
This is pretty cool idea, but it segfaults on my system when I try to search my code directory. Have you tried it on any large corpus of data?

acconsta:
Thanks for reviewing. I think I solved the problem — I was overrunning the file buffer because I thought it's null terminated.

I tried a randomized testing approach on a large corpus, and I haven't managed to crash it since (except for unrelated issue #741). I encourage you to take another look.

ggreer:
Thanks a lot for this PR. You've improved the speed of ag more than I've managed to in the past two years!

Issue #734, opened 2015-08-23
Limit Number of Results


xanderdunn:
It would be really great to be able to enforce a max number of results.  This is largely to prevent a never-ending search executed in some directory.  For my use of ag, I'm never going to be interested in more than 1000 results.  As a matter of fact, I doubt I'd ever even look at anything with ~100 results.  Sometimes searches will take a very long time and return hundreds of thousands of results, whereas I would have been happy with just seeing the first 100 results which would be enough to tell me that this is not the search I wanted to do.

sach1t:
I too have that problem.

codekiln:
+1

PR #738, opened 2015-09-01
Add '-j/--max-matches NUM' option

sach1t:
Allow the search to stop after NUM matches have been found.
Useful for large code bases where finding all matches of a pattern
may take too long or be unimportant.

Introduce counter for the number of matches. After NUM matches are
found stop adding files to work queue and stop worker threads.

When used with --stat the number of matches is reported to be
consistent with NUM despite there possibly being more matches
that were found.

Resolves issue #734.

### DIFF
--- a/doc/ag.1
+++ b/doc/ag.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "AG" "1" "February 2015" "" ""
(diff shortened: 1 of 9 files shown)


laktak:
@ggreer could this PR be merged? thanks!

baopham:
There are conflicts now but this would be really useful. @sach1t any plan updating the PR?

sach1t:
I wouldn't mind updating the PR, but it would not be worth the effort
unless there is a good chance that it will be considered for getting merged
this time.

PR #741, opened 2015-09-07
Fix invalid memory access

acconsta:
I've noticed occasional segfaults from accessing beyond the end of
the file buffer. This adds the correct bound to index variable i.

### DIFF
--- a/src/print.c
+++ b/src/print.c
@@ -133,7 +133,7 @@ void print_file_matches(const char *path, const char *buf, const size_t buf_len,
         }
 
         /* We found the end of a line. */
-        if (buf[i] == '\n' && opts.before > 0) {
+        if ((i == buf_len || buf[i] == '\n') && opts.before > 0) {
             if (context_prev_lines[last_prev_line] != NULL) {
                 free(context_prev_lines[last_prev_line]);
             }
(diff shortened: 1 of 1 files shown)


ggreer:
Can you create a test case or give me an example file & query that this fails on? I haven't been able to reproduce it.

Also, your change causes matches at the very end of a file to not be printed. I've added a test case to master that causes this PR to fail. If you rebase off master, `make test` should fail for you.

acconsta:
Debugging this was a throwback to my operating systems class.

It's pretty easy to verify (by printing or asserting) that `print_file_matches` will read one past the buffer end (address `buf[buf_len]`) on line 136.

Interestingly, that's **rarely** a problem (even though it's a buffer overrun) because of how `mmap` works — mmap buffers are implicitly rounded up to the nearest multiple of 4096 bytes. But *if* `buf_len` is already a multiple of 4096 bytes, and *if* `buf[buf_len]` isn't the start of another mmapped file, you get a segfault. 

I can reproduce this reliably under Linux 3.16.0 by searching very large codebases e.g. https://github.com/torvalds/linux with random strings. Specifically, the pattern `"W*"` seems to segfault every time on that repo.

acconsta:
I just repushed the branch with a fix for the end-of-file problem. Take another look (at your convenience).

ggreer:
Excellent work. Thanks so much for tracking this down.

Issue #744, opened 2015-09-08
-G hides smartcase


albfan:
Is supposed that smartcase works by default.

>   -S --smart-case
>              Match case-sensitively if there are any uppercase letters in PATTERN, case-insensitively otherwise. Enabled by default.

tested on ag 0.31.0

<pre>
$ ag --version
ag version 0.31.0

Features:
  +jit +lzma +zlib
</pre>

a test case:
tested on https://github.com/albfan/ag.vim

<pre>
$ ag --group --column -Gvim$ groupfile
alberto@archlinux-alberto:~/.vim/bundle/ag.vim
$ ag --group --column -S -Gvim$ groupfile
plugin/ag.vim
5:42:command! -count -nargs=*               AgGroupFile call ag#AgGroup(<count>, 0, <f-args>)
</pre>

PR #758, opened 2015-09-25
Don't alter casing if the file pattern is lowercase

Arkanosis:
Closes ggreer/the_silver_searcher#744.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -556,10 +556,7 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
 
     if (file_search_regex) {
         int pcre_opts = 0;
-        if (opts.casing == CASE_SMART) {
-            opts.casing = is_lowercase(file_search_regex) ? CASE_INSENSITIVE : CASE_SENSITIVE;
-        }
-        if (opts.casing == CASE_INSENSITIVE) {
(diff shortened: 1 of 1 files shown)


Arkanosis:
Anything I can do to make testing / merging easier?

Maybe with a more complete description of the bug:

`-S` / `--smart-case` is expected to make ag case-insensitive for regexps that are fully lowercase, and only in this case. Currently, if the *file* pattern (ie. with `-G`) is lowercase, ag becomes case-insensitive not only for the *file* pattern, but for the *search* pattern as well. ag should be case-insensitive for the *search* pattern only if the *search* pattern itself is lowercase.

The proposed changeset fixes this side-effect.

Side note: in my opinion, having distinct smart case flags for the *file* pattern and the *search* pattern would be better: there's no reason for the user to systematically want the same matching mode for both patterns.

Best regards,

ggreer:
What happens if I run `ag -i 'BLAH.*'` ?

Arkanosis:
Indeed, some stupid mistake on my part. It does run case-insensitive search on the file content, but not on the file name. I did not intend to change this behavior. The updated PR fixes this and passes the new test case. Thanks.

It stills bugs me to have the same `-i` / `-s` / `-S` flags to control the case-sensitivity of both the files names and the files contents, though, but that could be the subject of another PR.

ggreer:
Thanks for being patient.

I'm not super-concerned about those flags applying to both text and filename search. I think the behavior you want is best served by chaining invocations.

PR #760, opened 2015-10-01
Add TOML filetype support

natemara:
TOML is a markup language used for the configuration of Rust's main build tool, [Cargo](https://crates.io). The spec is online [here](https://github.com/toml-lang/toml). This would allow Ag to search through TOML files via its regex file rules.

PR #761, opened 2015-10-01
Fix list_file_types test

natemara:
My addition of the TOML regex in #760 broke the test that listed all of the supported languages. This should fix that.


natemara:
I'll fix this mess back in #760

Issue #697, opened 2015-06-19
List regex flavor options in README


tecfu:
What families of regular expressions is Ag compatible with? i.e. POSIX, POSIX Extended, Vim, Perl...

This isn't immediately clear...there's no definitive answer online, the man page says nothing on the subject. If stated in the README and it would save new users lots of time and confusion and give instant clarity to new users.

The only answer I've found is on stack overflow and comes from a non-definitive source. That answer states says Ag uses PCRE, 

http://stackoverflow.com/questions/22190334/ag-silver-searcher-what-are-the-rules-for-patterns

but that's clearly not always the case, as I can use POSIX character classes from my terminal like so

```
ag [[:digit:]] src/*.js
```

decaff:
From looking at the code...

Ag uses the pcre(3) library with the PCRE_MULTILINE option enabled.  Regex features are listed here:       

````
man 3 pcrepattern
````

decaff:
A man page update is available in [this commit](https://github.com/decaff/the_silver_searcher/commit/f45828812cfe82c3e2dce9fd8c0297fab54de3a6) .

tecfu:
Great work! I hope to see that get merged.

tgroshon:
Did a PR get submitted or merged for this?

decaff:
No.  Geoff is pretty busy.  I'm letting my commits accumulate.  I'm sure it will be sorted when he has some free time.

PR #762, opened 2015-10-01
Mention that Ag uses PCRE

gwerbin:
I wanted to know what kinds of expressions I could use, but it didn't seem to be documented (anywhere other than [StackOverflow](http://stackoverflow.com/a/22191815)). Seems reasonable that it should be in the official docs.

### DIFF
--- a/doc/ag.1.md
+++ b/doc/ag.1.md
@@ -7,7 +7,7 @@ ag(1) -- The Silver Searcher. Like ack, but faster.
 
 ## DESCRIPTION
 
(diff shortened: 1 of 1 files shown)


gwerbin:
I guess this is a duplicate of https://github.com/ggreer/the_silver_searcher/issues/697 and the corresponding commit https://github.com/decaff/the_silver_searcher/commit/f45828812cfe82c3e2dce9fd8c0297fab54de3a6

PR #772, opened 2015-10-21
Adds restructuredText (.rst) language type

moreati:
[restructuredText](https://en.wikipedia.org/wiki/ReStructuredText) is a lightweight markup format used widely for e.g.

 - Python documentation
 - Static site generators
 - Academic papers


ggreer:
Seems like a rather long name, but ok.

moreati:
I debated calling the option `--rst`, and I'm happy to change it if you prefer

PR #774, opened 2015-10-25
Adding CMake build

coreylowman:
Added a CMakeLists.txt file to enable building with cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,63 @@
+cmake_minimum_required(VERSION 2.8)
+
+project(AG)
+
+#get the version from configure.ac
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
+cmake_minimum_required(VERSION 2.8)
+
+project(AG)
+
+#get the version from configure.ac
+set(EXECUTE_COMMAND grep -Po "(?<=\\[)([0-9.]+.[0-9]+.[0-9]+)(?=\\])" ${AG_SOURCE_DIR}/configure.ac)
+execute_process(COMMAND ${EXECUTE_COMMAND} OUTPUT_VARIABLE AG_VERSION)
```
rphillips: it would be portable to use cmake's file read and cmake's string match support to pull out the version.

PR #781, opened 2015-10-30
Fix --word-regexp not applying to alternates

danielshahaf:
`ag -w 'foo|bar'` would match `foobar`.  Test included.


ggreer:
I thought you were using a capture, but I didn't understand what the `:` was for. I dug around in the PCRE docs and apparently, `(?:...)` means "non-capturing group". Since captures hurt performance, your fix is exactly what I want. Thanks!

PR #787, opened 2015-11-04
Added support for user-defined languages 

ban-dana:
Language definitions are read from ~/.aglang file on startup.
The format follows:
# Lines starting with a # symbol on position 0 are skiped (comments)
# next line will define a new language lan with etensions .lll and .lan
lan : .lll, .lan

# next line will add a custom extension .ccc to cpp language defintion
.cpp + .ccc

### DIFF
--- a/src/lang.c
+++ b/src/lang.c
@@ -4,6 +4,10 @@
 #include "lang.h"
 #include "util.h"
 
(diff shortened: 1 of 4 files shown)


elektronik2k5:
Awesome, thank you @ban-dana for this!

PR #788, opened 2015-11-06
Fix symlink support on Win32

ban-dana:


ggreer:
Thanks! I hardly use Windows, so I'm trusting you that this works. :)

PR #798, opened 2015-11-27
Support TypeScript with --ts (.ts, .tsx)

chuckd:
Could also be --typescript but given JavaScript is in as --ts I thought I'd go with that.


ggreer:
Thanks!

indiesquidge:
Hey @ggreer. Any ETA on the next release that will include this option? I can't seem to install `ag` manually so I've been using Homebrew's version.

PR #805, opened 2015-12-07
Match same-case non-ascii when doing strncasestr.

### LINKED ISSUE #241: Case insensitive search fails to find special chars (german umlaut)
Given a file with the text `Foobarä`, `ag -i foobar` matches, but `ag -i foobarä` does not.

(`ag Foobarä` matches, too)

This is with a utf8 console and text file, using current master.

`ag -i 'foobarä.*'` also matches, so there appears to be some optimization regarding fixed pattern lengths, which causes this.

wapiflapi:
This doesn't fix the problem that we don't know how to do non-case-sensitive matching on UTF8 and the like, but at least this makes it so we don't fail when it's an exact match.

Fixes #241 (or at least strictly improves.)


wapiflapi:
Test cases:
 - [x] `echo café | ag -i café` works now (improvement).
 - [x] `echo CAFÉ | ag -i CAFÉ` works now (improvement).
 - [ ] `echo CAFÉ | ag -i café` still doesn't work (no regression).
 - [ ] `echo café | ag -i CAFÉ` still doesn't work (no regression).

ggreer:
Umm... the first two test cases work for me with the latest version of Ag. What platform are you on? Also, what is the output of `locale`? For me, it's: 

    LANG="en_US.UTF-8"
    LC_COLLATE="en_US.UTF-8"
    LC_CTYPE="en_US.UTF-8"
    LC_MESSAGES="en_US.UTF-8"
    LC_MONETARY="en_US.UTF-8"
    LC_NUMERIC="en_US.UTF-8"
    LC_TIME="en_US.UTF-8"
    LC_ALL=

wapiflapi:
I have basically the same locale you do:
```
LANG=en_US.UTF-8
LANGUAGE=
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL=
```

Testing with latest master from git, `ag version 0.31.0` on `x86_64 / Ubuntu LTS`.

For the record, here is the md5sum of the input, maybe that's encoded differently for you or something? I don't really buy this but just checking.
```
$ echo café | md5sum
6e99834b7c3e3fd53529a5489725d7e8  -
$ echo café | ./ag café
café
$ echo café | ./ag -i café # No match
```

PR #619, opened 2015-02-24
lang: Allow both extensions and full filenames

mingwandroid:
Because many filetypes (e.g. make) are not indicated by the
file extension but instead by the full filename (Makefile).

Renames extension to pattern, but retains the existing format
where each pattern is treated as an extension. Full filenames
are indicated by prefixing the pattern with a '^'.

PR #810, opened 2015-12-16
Add support for non-extension filenames and use them for scons, cmake and make

bwalle:
This patch adds support for languages that have files without an extension.

The next patches utilize that new support.

### DIFF
--- a/src/lang.c
+++ b/src/lang.c
@@ -5,132 +5,145 @@
 #include "util.h"
 
 lang_spec_t langs[] = {
-    { "actionscript", { "as", "mxml" } },
-    { "ada", { "ada", "adb", "ads" } },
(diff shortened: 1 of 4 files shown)


kastiglione:
I'd like this! I don't know the codebase, but it seems reasonable. The CI failure is code formatting. Are pull requests being committed? The number of open PRs makes it seem like maybe they're not.

mingwandroid:
I have an earlier PR that does the same thing but much less intrusively:

https://github.com/ggreer/the_silver_searcher/pull/619

nyxus:
I would really like it if the PR gets accepted into to master. Often I come across big codes bases with a lot of cmake. Having this feature means I can also us ag for finding stuff in those files!

PR #822, opened 2016-01-12
Fix filetype option

y-yamagata:
In the case of next, all files are searched.
```
$ ls ag_test
hoge    hoge.as hoge.c
$ ag hoge --actionscript ag_test
ag_test/hoge.as
1:hoge

ag_test/hoge.c
1:hoge

ag_test/hoge
1:hoge
```

expect:
```
$ ag hoge --actionscript ag_test
ag_test/hoge.as
1:hoge
```


ggreer:
Oops! Thanks for fixing this.

Issue #626, opened 2015-03-07
Feature request: add -x/--line-regexp (match whole line) option


okdana:
I've already submitted a pull request for this (#625), but i thought i ought to create an issue as well so that people can find it more easily.

Summary: Requesting an `ag -x` option equivalent to `grep -x`, which forces the pattern to match only against whole lines

ggreer:
I'll leave your issues open, but in the future, please don't create issues for each pull request. Over 200 people get emailed for each issue or PR, and searching the whole project isn't difficult.

okdana:
Oh, i'm sorry. I'm used to needing a ticket to document every change, i don't have much experience with GitHub etiquette. Sorry again for the spam!

ggreer:
Yeah, I figured it was an honest mistake. Don't feel bad about it.

cup:
@ggreer any comment on the actual issue? `-x` is not just defined by GNU, but POSIX as well:

http://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html

granted, Ag is not a POSIX tool, but its worth mentioning

PR #824, opened 2016-01-14
Fix literal word-regexp.

y-yamagata:
```
$ cat blah.txt
abcblah blah blah

# Match. (GNU grep 2.6.3)
$ grep -wF 'blah blah' blah.txt
abcblah blah blah

# No match.
$ ag -wF 'blah blah' blah.txt

# expect?
$ ag -wF 'blah blah' blah.txt
1:abcblah blah blah
```

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -66,8 +66,8 @@ void search_buf(const char *buf, const size_t buf_len,
                     /* It's a match */
                 } else {
                     /* It's not a match */
-                    match_ptr += opts.query_len;
-                    buf_offset = end - buf;
+                    match_ptr += find_skip_lookup[0] - opts.query_len + 1;
+                    buf_offset = match_ptr - buf;
                     continue;
                 }
             }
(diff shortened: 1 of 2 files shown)


ggreer:
Nice catch, and nice test cases. Thank you very much.

ldong:
Follow up question:

For example, I want to search exact word **577**

File content

```
577
577.1
5777.4
```

I tried to use `ag -wF "577"`, no luck and it keeps appearing other words as well. 

```
1:577
2:577.1
```
I only want to show `1:577`. Please help, thanks.

okdana:
The purpose of the `-w` option is to match 'whole words', where a word is a string beginning and ending with a word character `[A-Za-z0-9_]`, and bounded on each side by either the beginning/end of line or a non-word character. In your example, the `577` in `577.1` is bounded on the left side by the beginning of line and on the right side by the non-word character `.` — so `-w` is behaving the way i'd expect it to.

What you might be expecting is a definition of 'word' where it's a string bounded by either the BOL/EOL or a *space*. In non-fixed-string mode you could achieve that with the pattern `(^|\s)577(\s|$)` — but there's no way to get that precise with `-F`.

Alternatively, you might be interested in a 'whole line' match. `grep` provides this with `-x`. Unfortunately, `ag` doesn't currently have that option — see #626.

PR #828, opened 2016-01-25
Support WiX files with --wix

sebgod:
Add the XML source [file extensions](http://wixtoolset.org/documentation/manual/v3/overview/files.html)
that are commonly used by WiX installer projects.


sebgod:
@jschpp here with the correct file extensions for WiX

ggreer:
Thanks!

PR #831, opened 2016-01-29
Made --after and --before have required arguments

### LINKED ISSUE #778: Long version of option -A (--after) doesn't work
```
alex@bender > /tmp/the_silver_searcher-0.31.0 $ ag rest ./doc
doc/ag.1
131:Skip the rest of a file after NUM matches\. Default is 0, which never skips\.
186:\fB\-u \-\-unrestricted\fR
229:It is possible to restrict the types of files searched\. For example, passing \fB\-\-html\fR as the \fBfile\-types\fR parameter will search only files with the extensions \fBhtm\fR, \fBhtml\fR, \fBshtml\fR or \fBxhtml\fR\. For a list of supported \fBfile\-types\fR run \fBag \-\-list\-file\-types\fR\.
```
```
alex@bender > /tmp/the_silver_searcher-0.31.0 > ag --after 4 rest ./doc
ERR: Error stat()ing: rest
ERR: Error opening directory rest: No such file or directory
doc/ag.1
51:Color codes for result match numbers\. Defaults to 30;43\.
52-.
53-.TP
```
```
alex@bender > /tmp/the_silver_searcher-0.31.0 $ ag -A4 rest ./doc
doc/ag.1
131:Skip the rest of a file after NUM matches\. Default is 0, which never skips\.
132-.
133-.TP
134-\fB\-\-[no]numbers\fR
135-Print line numbers\. Default is to omit line numbers when searching streams\.
186:\fB\-u \-\-unrestricted\fR
187-Search \fIall\fR files\. This ignores \.agignore, \.gitignore, etc\. It searches binary and hidden files as well\.
188-.
189-.TP
190-\fB\-U \-\-skip\-vcs\-ignores\fR
229:It is possible to restrict the types of files searched\. For example, passing \fB\-\-html\fR as the \fBfile\-types\fR parameter will search only files with the extensions \fBhtm\fR, \fBhtml\fR, \fBshtml\fR or \fBxhtml\fR\. For a list of supported \fBfile\-types\fR run \fBag \-\-list\-file\-types\fR\.
230-.
231-.SH "IGNORING FILES"
232-By default, ag will ignore files whose names match patterns in \.gitignore, \.hgignore, or \.agignore\. These files can be anywhere in the directories being searched\. Ag also ignores files matched by the svn:ignore property if \fBsvn \-\-version\fR is 1\.6 or older\. Finally, ag looks in $HOME/\.agignore for ignore patterns\. Binary files are ignored by default as well\.
233-.
```
```
$ ag -V
ag version 0.31.0

Features:
  +jit +lzma +zlib
$ uname -a
Linux bender 3.19.0-15-generic #15-Ubuntu SMP Thu Apr 16 23:32:37 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
```

Tarball have downloaded from http://geoff.greer.fm/ag/

aubreyand:
Still works without args
Fixes #778


ggreer:
Sorry, but I'd much rather have `--after` and `--before` with with optional args.

AubreyEAnderson:
I guess i didnt make this clear at all in the pull request. It still works with optional args, the equals sign is just no longer required. It seems that required argument doesn't actually mean that the argument is required

I'm not really sure why it is that way. My guess is that it has something to do with the way the the function looks for arguments to flags. That is, if the argument is optional it looks for something after an equals sign, but if the argument is required it looks for it as the next value in argv unless the next value is another flag.

ggreer:
Ohhhh. Whoops.

timvisher:
i think the gnu standard is to have long options require a `=` to be set and short options to be whitespace delimited. Why would we want to have whitespace delimited long options?

adius:
More convenient to type, easier to read and most important users will expect it to work!

olejorgenb:
> i think the gnu standard is to have long options require a = to be set and short options to be whitespace delimited. Why would we want to have whitespace delimited long options?

No: `grep PATTERN --context 3 FILE` works. `ls --sort time` works.

If this PR isn't merged, can the manual at least be updated to not be misleading? 

(the zsh completion provided by https://github.com/zsh-users/zsh-completions is also wrong regarding --context, etc.)

Issue #755, opened 2015-09-17
.aginore not ignoring filenames containing several dots


thranur:
If my `.agignore` file contains:

    *.htm

And if I have two files `a.htm` and `a.b.htm`, the latter is **not** ignored by `.agignore`.

Here's an example:

    $ echo "needle" > haystack.htm
    $ echo "needle" > haystack.a.htm
    $ echo "*.htm" > .agignore
    $ ag needle
    haystack.a.htm
    1:needle

I found out that using `**.htm` works in this case, but if this is intentional, it is very uninituitive, especially since it does not follow `.gitignore` syntax. I'd recommend in this case mentioning this explicitly in https://github.com/ggreer/the_silver_searcher/wiki/Advanced-Usage, which says that:
> .agignore uses the same patterns as .gitignore and .hgignore.

JIghtuse:
It works on 0.30.0. git bisect told me it was intentional change indeed (introduced in d3a5de43e4c7b8735fede1c). I don't quite get why ag should behave like this, but fix is trivial - revert commit to use strrchr instead of strchr. I can sent PR if @ggreer agrees to accept it.

epmatsw:
Ugh, that's my bad on that one. The issue that was fixed was that you couldn't ignore `*.min.js` files, because it would just get the `.js`, which doesn't match any pattern. It seems like the fix is going to have to be more nuanced than my change :(

fuzzy76:
As a web developer, not being able to ignore .min.js files means it is completely broken :(

brianpeiris:
FYI, the workaround for this is to add `*.a.htm` or `*.*.htm` to your `.agignore`.

fuzzy76:
Why aren't you comparing the entire filename when matching against wildcards? I'm pretty sure that's what git does when parsing ```.gitignore```

brianpeiris:
@fuzzy76 It's just a bug but there are performance concerns that make the fix a bit complicated. @ggreer mentions it in his blog post: http://geoff.greer.fm/2014/10/13/help-me-get-to-ag-10/

jean:
Argh. Kind of amazed to realize that glob matching (e.g. `*.min.*`) is "special" in `ag` ..

denilsonsa:
I believe this issue was fixed by 221e4c2f070cb69825a42bd43a6bd79e18c0729f

spookylukey:
[...]

PR #834, opened 2016-02-01
Add failing test for ignoring files with multiple extensions

brianpeiris:
Covers the issue reported in #755
The `.min.js` files should be ignored due to the general `*.js` rule.


ggreer:
Please put this in `tests/fail/` so it doesn't break Travis-CI

brianpeiris:
This test should normally pass. I just added it for whoever wants to tackle the fix. Should it still go into `tests/fail/`?

Issue #785, opened 2015-11-03
Add zcml and xml extension for --plone


rodfersou:
Plone has some configuration files with `zcml` extension, and others with `xml` extensions that should be included into `ag` search.

jschpp:
Hey @ggreer could you still have a look at my pr?

PR #839, opened 2016-02-14
Added filetypes to test case

jschpp:
Test was failing. Now it shouldn't


jschpp:
fixes #786 and should close #785

rodfersou:
thanks @jschpp

ggreer:
I'll defer to you Plone users, but searching .py and .xml seem like it'd cause tons of false positives.

rodfersou:
@ggreer in our codebase it is normal to search into ``.xml`` files where we create profile configuration.

In Plone we just use ``.xml`` files in this case.

PR #841, opened 2016-02-17
Makefile.w32: Remove Perl-regex grep dependency

SebastianGodelet:
msys grep does not support the -Po option, so use sed instead
README.md: better describe that building requires an msys/MinGW
environment


ggreer:
Thanks. I don't use Windows, so I'm trusting that this works.

SebastianGodelet:
@ggreer , I've tested this with the old msys and the new msys2 for both 32-bit and 64-bit, it seems to work fine. Unfortunately Travis CI does not support pure Windows builds.
When I've time :tm: then I'll investigate using the GCC cross compiler that exists in any modern Linux distribution, thus might also work with Travis CI.

PR #842, opened 2016-02-17
On OpenBSD use pledge to drop privileges.

n0la:
With suggestions by Doug Hogan and the initial work by Michael McConville.

OpenBSD has added a new interface: [pledge](http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man2/pledge.2) which allows an application to drop privileges that are not needed. As `ag` is an application that searches a whole variety of files this should improve security, when unleashing `ag` on a set of unknown files.


ggreer:
Ooh! Thanks!

PR #846, opened 2016-02-20
Fix #845

### LINKED ISSUE #845: Duplicate but conflicting definition of enum ag_compression_type
Hello,

Our compiler goto-cc reported the following conflict:

file src/decompress.c line 220: 
reason for conflict at #this: number of members is different (4/5)
```
enum  { AG_NO_COMPRESSION=0, AG_GZIP=1, AG_COMPRESS=2, AG_ZIP=3 }
enum  { AG_NO_COMPRESSION=0, AG_GZIP=1, AG_COMPRESS=2, AG_ZIP=3, AG_XZ=4 }
error: conflicting function declarations 'is_zipped'
old definition in module 'search' file src/util.h line 78
enum  (const void *, const signed int)
new definition in module `decompress' file src/decompress.c line 220
enum  (const void *buf, const signed int buf_len)
```

This is due to definitions of ag_compression_type both in src/util.h as well as src/decompress.h, where the one in util.h is missing AG_XZ.

Best,
Michael

jschpp:
Fix for #845
I don't get why `decompress` and `ag_compression_type` were declared in util.h
If there was a logic behind it simply ignore this^^

### DIFF
--- a/src/search.h
+++ b/src/search.h
@@ -28,6 +28,7 @@
 #include "options.h"
 #include "print.h"
 #include "util.h"
+#include "decompress.h"
 #include "uthash.h"
 
 size_t alpha_skip_lookup[256];
--- a/src/util.h
+++ b/src/util.h
@@ -42,12 +42,6 @@ typedef struct {
     struct timeval time_end;
 } ag_stats;
 
-typedef enum {
-    AG_NO_COMPRESSION,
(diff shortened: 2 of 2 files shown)


ggreer:
I forget. I think it was a mistake. Thanks for noticing this redundancy.

Issue #716, opened 2015-07-17
default `--context` seems to be `0` and not `2` as stated in `ag --help`


eksperimental:
running released branch "ag version 0.30.0"

```bash
ag "you may not use"
LICENSE
4:  you may not use this file except in compliance with the License.
```

```bash
ag "you may not use" -C2
LICENSE
2-
3-  Licensed under the Apache License, Version 2.0 (the "License");
4:  you may not use this file except in compliance with the License.
5-  You may obtain a copy of the License at
6-
```
```bash
ag --version
ag version 0.30.0

Features:
  +jit +lzma +zlib
```

ggreer:
`ag -C` defaults to a context of 2.

eksperimental:
So why in the first examole there are is no context

eksperimental:
Sorry i see what u mean.

PR #848, opened 2016-02-22
Fix default value typo

thomasdavis:
When testing `ag` without -A and -B, I get 0 lines shown above and below. I believe that this is a typo.


ggreer:
I guess the docs aren't clear. If you specify -A or -B without [lines], you will get two lines of context above/below. This is related to #716.

thomasdavis:
Ahh I feel stupid but the other thread makes me feel better!

PR #857, opened 2016-02-28
Just a friendly reminder to use binary searching

br0ns:
`filename_filter` [searches a hardcoded list](https://github.com/ggreer/the_silver_searcher/blob/master/src/ignore.c#L358) of files to ignore.  At present a linear search is used, which is OK because the list is very short.  However, should the list grow in the future a binary search will be better suited.


ggreer:
Thanks, but I doubt that list will grow in the future. If it does, I'm sure I'll be cognizant enough to have a better search.

PR #860, opened 2016-03-05
Add support for .factor files.

erg:
https://factorcode.org


ggreer:
Huh, I'd never heard of that language before. Thanks for the PR.

PR #865, opened 2016-03-15
Skip one_device.t if /dev/shm isn't a mountpoint

### LINKED ISSUE #864: one_device.t fails if /dev/shm isn't a separate mount point
If `/dev/shm/` isn't a mountpoint, `one_device.t` fails because it (correctly) descends into `/dev/shm`. This is notably the case on Cygwin, but I believe there's nothing that strictly requires `/dev/shm` to be a tmpfs partition.

me-and:
If /dev/shm isn't a separate device to its parent directory, skip using
it to test limiting a search to a single device.  If /dev/shm isn't a
mountpoint there's no obvious alternative for checking Ag doesn't
descend into one, so just skip the test entirely.

Fixes #864.

### DIFF
--- a/tests/one_device.t
+++ b/tests/one_device.t
@@ -4,6 +4,9 @@ Setup:
   > if [ ! -e "/dev/shm" ]; then
   > echo "No /dev/shm. Skipping test."
   > exit 80
+  > elif [ "$(stat -c%d /dev/)" == "$(stat -c%d /dev/shm/)" ]; then
+  > echo "/dev/shm not a different device.  Skipping test."
+  > exit 80
   > fi
   $ TEST_TMPDIR=`mktemp -d --tmpdir=/dev/shm ag_test.XXX`
   $ echo "blah" > $TEST_TMPDIR/blah.txt
--- a/tests/one_device.t
+++ b/tests/one_device.t
@@ -4,7 +4,7 @@ Setup:
   > if [ ! -e "/dev/shm" ]; then
   > echo "No /dev/shm. Skipping test."
   > exit 80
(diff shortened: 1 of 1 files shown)


me-and:
Hmm. The Travis CI build skipped the test, which I wasn't expecting. Presumably checking the contents of `/etc/mtab` doesn't work in the Travis build environment. Let's try finding an alternative…

me-and:
Okay, now I'm just baffled. The Travis test is failing with the below error:

```
tests/one_device.t: failed
--- tests/one_device.t
+++ tests/one_device.t.err
@@ -8,6 +8,7 @@
   > echo "/dev/shm not a different device.  Skipping test."
   > exit 80
   > fi
+  /bin/sh: 6: [: 5: unexpected operator
   $ TEST_TMPDIR=`mktemp -d --tmpdir=/dev/shm ag_test.XXX`
   $ echo "blah" > $TEST_TMPDIR/blah.txt
   $ ln -s $TEST_TMPDIR other_device
```

I don't see this on my CentOS box or on Cygwin, and looking at the code I really can't see why `[` is an "unexpected operator" at this point; it's immediately after an `elif` and has similar syntax to the preceding `if` statement which predates my changes.

me-and:
Sussed it – Travis's shell doesn't like `[[ ... ]]` tests, and is evidently running a shell that's pretty strict about its POSIX compliance, meaning using `==` within a `[ ... ]` test isn't permitted: the test needs to use `=` instead.

ggreer:
Nice detective work. Thanks for figuring this out and submitting a fix.

PR #867, opened 2016-03-15
Use printf in place of echo in tests

### LINKED ISSUE #866: column.t and ignore_extensions.t fail in CentOS 6 and Cygwin
On both my Cygwin and CentOS 6 boxes, `column.t` and `ignore_extensions.t` fail. I'm bunching them both together as the failure appears to be something to do with `[1]` output, which I don't (currently) understand; it may be that they're separate problems.

```
tests/column.t: failed
--- column.t
+++ column.t.err
@@ -6,13 +6,10 @@
 Ensure column is correct:

   $ ag --column "blah\nb"
-  blah.txt:1:1:blah
-  blah.txt:2:0:blah2
+  [1]

 # Test ackmate output. Not quite right, but at least offsets are in the
 # ballpark instead of being 9 quintillion

   $ ag --ackmate "blah\nb"
-  :blah.txt
-  1;blah
-  2;0 6:blah2
+  [1]
```

```
tests/ignore_extensions.t: failed
--- ignore_extensions.t
+++ ignore_extensions.t.err
@@ -19,7 +19,7 @@
 Ignore patterns with multiple extensions in root directory:

   $ ag "targetB"
-  [1]
+  aFile.test.txt:1:targetB

 Do not ignore patterns with partial extensions in root directory:

@@ -34,7 +34,7 @@
 Ignore patterns with multiple extensions in subdirectory:

   $ ag "targetE"
-  [1]
+  subdir/anotherFile.test.txt:1:targetE

 Do not ignore patterns with partial extensions in subdirectory:

```

me-and:
The behaviour of echo is inconsistent between implementations, in
particular regarding whether it expands escapes like `\n`.  Use printf
consistently in test scripts to avoid test failures from inconsistent
echo behaviour.

Fixes #866.

### DIFF
--- a/tests/adjacent_matches.t
+++ b/tests/adjacent_matches.t
@@ -2,7 +2,7 @@ Setup:
 
   $ . $TESTDIR/setup.sh
   $ alias ag="$TESTDIR/../ag --workers=1 --parallel --color"
-  $ echo blahfoofooblah > ./fooblah.txt
+  $ printf 'blahfoofooblah\n' > ./fooblah.txt
 
 Highlights are adjacent:
 
--- a/tests/case_sensitivity.t
+++ b/tests/case_sensitivity.t
@@ -1,8 +1,8 @@
 Setup:
 
   $ . $TESTDIR/setup.sh
-  $ echo Foo >> ./sample
-  $ echo bar >> ./sample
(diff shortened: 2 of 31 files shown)


ggreer:
Thanks for figuring this out and fixing it.

PR #871, opened 2016-03-22
Add YANG (RFC 6020) and Lux (github.com/hawk/lux) filetypes

chilikk:


avtobiff:
Note that #912 also adds support for YANG, but omits `.yin` files.

PR #887, opened 2016-04-26
Fix potential integer overflow in binary search

sach1t:


ggreer:
Heh, I was wondering when someone would find that line. Nice. 👍

PR #894, opened 2016-05-07
Fix #893 - Use [ \t] instead of \s (BSD grep)

### LINKED ISSUE #893: filetype.t test fails on FreeBSD
FreeBSD grep -E doesn't support `\s`.  Instead, use `[ \t]`.

cemeyer:


ggreer:
Thanks!

PR #895, opened 2016-05-08
Fix #347 - Keep enough context around to print in stream mode

### LINKED ISSUE #347: context line printing wrong when reading from a pipe
Tested with both ag 0.14 and ag 0.19.2 on OS X Mountain Lion & Mavericks

It's a somewhat understandable behavior, but other related tools (grep) do this correctly.

``` bash
# test setup
$ for f in a b c; do echo $f; done > test.txt
$ cat test.txt
a
b
c

# correct:
$ ag -B1 b test.txt
1-a
2:b

# correct:
$ ag -C1 b test.txt
1-a
2:b
3-c


# bug in -B handling:
$ cat test.txt | ag -B1 b
b

# bug in -C handling:
$ cat test.txt | ag -C1 b
b
:
```

(yes, that's a weird ":" printed at the end)

cemeyer:
Includes a new set of tests from #347, `pipecontext.t`.

### DIFF
--- a/src/print.c
+++ b/src/print.c
@@ -8,6 +8,7 @@
 #include "log.h"
 #include "options.h"
 #include "print.h"
+#include "search.h"
 #include "util.h"
 #ifdef _WIN32
 #define fprintf(...) fprintf_w32(__VA_ARGS__)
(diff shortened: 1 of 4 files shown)


ggreer:
Oh hecks yes! Thank you so much for this PR!

cemeyer:
Glad it helps!  I use `ag` basically every day.

forgottenswitch:
https://github.com/ggreer/the_silver_searcher/compare/master...forgottenswitch:ABC?expand=1 :)

forgottenswitch:
BTW, the implementations give different results:
(Don't know how it differs from piping (with it results are equal))
```
[the_silver_searcher]$ ./ag --nocolor --number '.' <( printf '\nfoo\n\nar\n' )
1:
2:
4:o
5:
7:8:
10:
[the_silver_searcher]$ ./ag --nocolor --number '.' <( printf '\nfoo\n\nar\n' )
1:
2:foo
3:
4:ar
[the_silver_searcher]$
```

PR #903, opened 2016-05-15
Improve multi-workers' performance by 15%

lxyscls:
Hi, Greer,
I have made some effort to improve the multiple workers' performance.
1. Set all the threads' affinities according to the CPU Threads in Round Robin manner.
2. Threads convert matches into thread specific dynamic string, so that workers can do more works in parallel.

Below are a few test charts I had made on my machines. The performance has been improved by 15%. The directory [linux-master](https://github.com/torvalds/linux) is about 720MB.
- ag --worker=$i -U -Q "#include" linux-master
  -  [Intel® Xeon® Processor E5-2648L](http://ark.intel.com/products/61426/Intel-Xeon-Processor-E5-2648L-20M-1_80-GHz-8_0-GTs-Intel-QPI) vs 16GB MEM
    image
  - [Intel® Xeon® Processor E5-2618L v3](http://ark.intel.com/products/83351/Intel-Xeon-Processor-E5-2618L-v3-20M-Cache-2_30-GHz?q=E5-2618L%20v3) \* 2 vs 128GB MEM
    image
- ag --worker=$i -U "#inc*" linux-master
  - [Intel® Xeon® Processor E5-2648L](http://ark.intel.com/products/61426/Intel-Xeon-Processor-E5-2648L-20M-1_80-GHz-8_0-GTs-Intel-QPI) vs 16GB MEM
    image
  - [Intel® Xeon® Processor E5-2618L v3](http://ark.intel.com/products/83351/Intel-Xeon-Processor-E5-2618L-v3-20M-Cache-2_30-GHz?q=E5-2618L%20v3) \* 2 vs 128GB MEM
    image

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -122,6 +122,9 @@ int main(int argc, char **argv) {
     if (pthread_mutex_init(&work_queue_mtx, NULL)) {
         die("pthread_mutex_init failed!");
     }
+    if (pthread_key_create(&worker_key, ag_specific_free)) {
+        die("pthread_mutex_init failed!");
+    }
 
     if (opts.casing == CASE_SMART) {
         opts.casing = is_lowercase(opts.query) ? CASE_INSENSITIVE : CASE_SENSITIVE;
--- a/src/main.c
+++ b/src/main.c
@@ -161,8 +161,10 @@ int main(int argc, char **argv) {
     }
 
     if (opts.search_stream) {
+        ag_setspecific();
         search_stream(stdin, "");
     } else {
+        ag_setspecific();
         set_affinity(pthread_self(), num_cores);
         for (i = 0; i < workers_len; i++) {
             workers[i].id = i;
--- a/src/main.c
+++ b/src/main.c
@@ -27,24 +27,24 @@ typedef struct {
 
 static void set_affinity(pthread_t tid, int num_cores) {
 #if defined(HAVE_PTHREAD_SETAFFINITY_NP) && defined(USE_CPU_SET)
-	if (opts.use_thread_affinity) {
-		static int cpu = 0;
-		cpu_set_t cpu_set;
-		CPU_ZERO(&cpu_set);
-		CPU_SET(cpu%num_cores, &cpu_set);
-		int rv = pthread_setaffinity_np(tid, sizeof(cpu_set), &cpu_set);
-		if (rv) {
-			log_err("Error in pthread_setaffinity_np(): %s", strerror(rv));
-			log_err("Performance may be affected. Use --noaffinity to suppress this message.");
-		} else {
-			log_debug("Thread %d set to CPU %d", tid, cpu);
-		}
-		cpu++;
-	} else {
-		log_debug("Thread affinity disabled.");
-	}
+    if (opts.use_thread_affinity) {
+        static int cpu = 0;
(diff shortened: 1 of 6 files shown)


bmalehorn:
I know this PR is rather old, but it seems interesting. Some questions:

1. Why did you move affinity code to its own function? It seems to be unchanged.
2. You buffer output per-thread until you reach the end of the file, right? Rather than buffering until completion of the entire thread.
3. To be clear, this only affects `ag` use cases with a lot of printing, right? In my experience, whenever `ag` prints out >100 hits, I usually kill the process with Ctrl-C and make my search more specific. This may not have been merged because it's a fairly aggressive refactor only to speed up a rare edge case.

lxyscls:
@bmalehorn 
1. I bind the main thread to a core.
2. Yes, but the buffer may grow very large.
3. Fast is fast, and I think the efficiency on cold data has more meanings.

forgottenswitch:
>> You buffer output per-thread until you reach the end of the file, right? Rather than buffering until completion of the entire thread.

> Yes, but the buffer may grow very large.

Buffer until size treshold?

forgottenswitch:
> This may not have been merged because it's a fairly aggressive refactor only to speed up a rare edge case.

If this is referring to the `printf` renames, then it could be avoided with `#define`s:
https://gist.github.com/forgottenswitch/bce1c206a1af8f937983adb0dfbee2dd

forgottenswitch:
The agressiveness could be dealt with `#define fprintf ag_fprintf` on top of `src/print.c`.
But it isn't a problem; the amount of code is.
Doing a lot of prints isn't a [frequent] usage patters, and is, therefore, not worth maintaining.

lxyscls:
@forgottenswitch
Because ag_fprintf can change his behavior during runtime, so I can't use #define.
Threads can output their matches to thread local buffer in parallel, and then print exclusively.

This solution can't resolve the cold data's inefficiency.

forgottenswitch:
OK. This worked for me: https://gist.github.com/forgottenswitch/397a47f4f4db6cca32277a04abef1912

As for the "cold data", the question is: does the speed up show up exclusively on large trees, and only when hitting a large number of matches? Because if yes, then what for to maintain such a thing?

forgottenswitch:
If there are other `fprintf`-s that should not be `ag_fprintf`-ed, then sorry, missed that.

lxyscls:
@forgottenswitch 
As for #define, maybe you are right. But I am not sure about other code paths, because as you know, it needs to call ag_setprint when it wants to print to the stdout.

With the cold data, my patch can't speed up anymore. Because the process needs to load all the files into memory, it is the main bottleneck. There is nothing to do with the number of files, loading a file into memory is more expensive than regex match.

forgottenswitch:
> But I am not sure about other code paths

OK, put a `/* In this file, wrap each fprintf into an ag_setprint */` above the defines :)

> With the cold data, my patch can't speed up anymore

But the charts are from cold data, aren't them?
Also, they are only for large number of matches (thousands).

For the pull to be considered, it should be shown that in more common case, that is, when the number of hits isn't too big (that is, when `ag "#include <alsa.h>" linux-master` is issued), the speed up of 15% does show up.

Otherwise, the sloc count is just not worth the hassle.

PR #907, opened 2016-05-19
Fix --files-with(out)-matches with single filename

### LINKED ISSUE #769: --files-with-matches and --files-without-matches don't work with single file on command line
```
$ git describe --tags 
0.31.0-21-gb0eb6d7
$ ./ag --version
ag version 0.31.0

Features:
  +jit +lzma +zlib
$ ./ag Silver NOTICE 
1:The Silver Searcher
$ ./ag --files-with-matches Silver NOTICE 
$ ./ag Gold NOTICE 
$ ./ag --files-without-matches Gold NOTICE 
$ 
```

ilmari:
print_path() by default doesn't print file names if there was only one
specified on the command line, but when listing files with or without
matches we always want to print it.

Also add a test for --files-with(out)-matches, of which there were none.

Fixes #769

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -424,6 +424,7 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
             case 'l':
                 needs_query = 0;
                 opts.print_filename_only = 1;
+                opts.print_path = PATH_PRINT_TOP;
                 break;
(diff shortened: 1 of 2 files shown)


ggreer:
Thanks for the test cases! ❤️

PR #910, opened 2016-05-23
Remove never-implemented --print-long-lines option.

pda:
The `--print-long-lines` option was never implemented, except in the never-merged and probably incomplete #707. Since the flag was added in TODO status, and later an implementation proposed, a `-W` / `--width` option was introduced in #720 which truncates the display of matches to the specified length.

I'd argue that `-W` makes `--print-long-lines` redundant; it solves the same problem (matches in very long lines such as minified JS dumping way too much text to the terminal), but in a way that doesn't prevent those matches from being reported to the user.

So, this PR proposes that the never-implemented stub of this option be removed.


pda:
Ping to remove CLI flag that does nothing.

DamienCassou:
There are conflicts

pda:
Conflicts resolved.

PR #913, opened 2016-05-29
remove shebang from bash completion script

mizuno-as:
now, bash completion script have shabang, but there is not executable.
Therefore, warning occurs in package build of Debian.

see: https://lintian.debian.org/tags/script-not-executable.html

I think that shebang (and executable flag) is unnecessary for bash completion script.

### DIFF
--- a/ag.bashcomp.sh
+++ b/ag.bashcomp.sh
@@ -1,5 +1,3 @@
-#!/bin/bash
-
 _ag() {
   local lngopt shtopt split=false
   local cur prev


ggreer:
I don't use bash completion (I use zsh), so I have no idea if your change breaks things. It seems to me like the safest change would be to make the script executable.

me-and:
There's no need for a shebang or for the script to be executable: Bash completion scripts are `source`d rather than being executed. See, for example, [the tar completion script in the core bash-completion project](https://github.com/scop/bash-completion/blob/master/completions/tar) or [Git's completion script](https://github.com/git/git/blob/master/contrib/completion/git-completion.bash).

mizuno-as:
I agree with me-and.
The bash completion script installed in Debian does not have any shebang and executable flag. I think that there is not the side effect of this commit.

PR #916, opened 2016-06-04
Add fflush call to print_file_matches

erantapaa:
When piping the output of `ag` into another program results are not seen by the other program until the output buffer is full.

This PR adds a `fflush` call to `print_file_matches` so that results are seen as they are generated.


ggreer:
Thanks!

PR #920, opened 2016-06-11
Fix some printf formatting specifiers.

Smattr:
Thanks for your work with this tool. I find it invaluable.

I accidentally stumbled across an issue with these format specifiers when applying [GCC's format attribute](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-g_t_0040code_007bformat_007d-function-attribute-3232) to the log functions. I can add another commit that makes this change as well if you like, but I wasn't sure if this might cause portability problems for some targets.


ggreer:
Sorry, but not all compilers support `%z` (eg: Visual Studio).

PR #924, opened 2016-06-24
Search files named Rakefile with --rake flag

ryangreenberg:
Ruby rake commands are typically kept in a file named "Rakefile", not "Rakefiles" (see examples at https://github.com/ruby/rake). This small change changes `--rake` to search the correct file name.


ggreer:
Whoops. Thanks!

PR #927, opened 2016-06-29
Update rpm link for RHEL7+

charlesrgould:


jschpp:
why not use `yum install epel-release.noarch`?

ggreer:
I went with #929 since it had more future-proof instructions. Thanks for the PR, though.

Issue #928, opened 2016-07-02
rpm link in RHEL7+ installation instructions return 404


davidcpell:
I'm working on installing `ag` on a Centos 7.2 machine, so I'm following the instructions for RHEL7+. When I try to run the first step:

`rpm -Uvh http://download.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-5.noarch.rpm`

I get a 404 error (same thing if I just paste the URL into a web browser).

Any recommendations?

jschpp:
try this instead `yum install epel-release.noarch`
then `yum install the_silver_searcher`

davidcpell:
Thanks, @jschpp. That worked! I've opened #929 to update the `README` if you don't mind giving that a look.

ggreer:
Fixed by #929.

PR #929, opened 2016-07-04
modify README with up-to-date installation instructions for redhat systems

davidcpell:
this is in regard to #928 

The instructions in the README don't seem to work anymore but @jschpp recommended installing the `epel-release.noarch` package before trying `the_silver_searcher`, which works (tested on Centos 7.2).


ggreer:
Thanks!

Issue #899, opened 2016-05-11
"ERR: Error: No query." or empty string


rlpowell:
The man page says:

       -l --files-with-matches
              Only print the names of files containing matches, not the matching lines. An empty query will print all files that would be searched.

but:

/bin/ag -l /home/rlpowell/src

doesn't work because that searches for the string /home/rlpowell/src in the current directory.

So attempting the obvious fix I get:

rlpowell@stodi> /bin/ag -l '' /home/rlpowell/src
ERR: Error: No query. What do you want to search for?

mattscamp:
I have noticed this as well. You can pass '.' as the query for a temporary workaround. I will see if I can take a look at this today.

PR #932, opened 2016-07-05
Accept empty query

mattscamp:
As reported in https://github.com/ggreer/the_silver_searcher/issues/899, when using the -l flag, in the docs it says that we should be able to pass an empty query, but isn't currently allowed.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -729,11 +729,12 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
     }
 
     if (accepts_query && argc > 0) {
-        // use the provided query
         if (!needs_query && strlen(argv[0]) == 0) {
-          opts.query = ag_strdup(".");
+            // use default query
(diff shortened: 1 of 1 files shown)


src/options.c
```diff
@@ -729,8 +729,13 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
     }
 
     if (accepts_query && argc > 0) {
-        // use the provided query
-        opts.query = ag_strdup(argv[0]);
+        if (!needs_query && strlen(argv[0]) == 0) {
+            // use default query
+            opts.query = ag_strdup(".");
```
bmalehorn: This won't match an empty file (e.g. `touch empty.txt`). You should use `.*` instead.

ggreer: It's the same query as a few lines down, which is exercised by `tests/empty_match.t`.

Issue #915, opened 2016-06-02
With * in gitignore, ag will get wrong result


gngshn:
Because there is some other git repo in one git repo, these is a pattern \* in the .gitignore file in deeply directories in our project. This lead ag don't search files in these directories.

```
|-a gitrepo dir
|----file1
|----file2
|----dir1
|----.git dir
|----.gitignore
|----deep dir
|----|----.gitignore with * pattern
|----|----gitrepo1 dir
|----|----gitrepo2 dir
|----|----....
```

Ag will ignore all gitrepo1/2/... dirs in deep dir
So, can ag ignore the '*' in .gitignore?
Ack-grep have ignored the \* in gitignore.

CleanCut:
I think I have a related issue.

I have a repository whose `.gitignore` looks like this.

```
*
!*/
!*.c
!*.cc
(etc.)
```

But ag ignores all files, even though the `.c` and `.cc` files were explicitly excluded from being ignored.

So I run `ag -u` and it takes forever, because it is searching through ~6GB of binary files produced by the project (which is why .gitignore is set up like it is).

Someone please let me know if I should make a separate issue for this.

msmith491:
I've also hit this issue.  Here is my .gitignore file:
```
*.swp
*
!/**/
!*.*
nimcache/
```
I use `ag` as the file aggregator for my Vim fuzzy searcher with `ag -g ""`, and with this .gitignore file I get nothing returned.

rudle:
https://github.com/Homebrew/brew/blob/master/.gitignore is an example of a gitignore file that will confuse The Silver Searcher

msmith491:
I've narrowed down to problem to here: https://github.com/ggreer/the_silver_searcher/blob/master/src/ignore.c#L333 .  It appears that `fnmatch` matches everything if a `*` character is provided and ag doesn't respect `.gitignore` includes after this.  Unfortunately this is a hot loop and my C-fu is weak enough that I'm not sure what I can change without severely affecting `ag`'s performance.

alecdwm:
@msmith491 With your changes, given the following fs structure:
```
.
./file
./dir
./dir/file
./dir/dir2
./dir/dir2/file
./.gitignore
./.git
```
With the following .gitignore:
```
*
!file
!dir
!dir/*
```
Git will see:
```
	new file:   dir/dir2/file
	new file:   dir/file
	new file:   file
```
But ag will only see:
```
dir/file
file
```

[...]

PR #940, opened 2016-07-15
Fixing bug #915

msmith491:
Wildcard ignores cause ag to discard all result files.  This commit
addresses this issue by first checking if a wildcard '*' gitignore is
present then performs a negative check against all files that would be
disqualified by the wildcard ignore with any negative-ignores
(represented by entries starting with '!' in a .gitignore file).  If a
negative match is found, the previous match against '*' is skipped.

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -1,6 +1,7 @@
 #include <ctype.h>
 #include <dirent.h>
 #include <limits.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -278,6 +279,7 @@ static int path_ignore_search(const ignores *ig, const char *path, const char *f
     char *temp;
     size_t i;
     int match_pos;
+    bool unignore = false;
 
     match_pos = binary_search(filename, ig->names, 0, ig->names_len);
     if (match_pos >= 0) {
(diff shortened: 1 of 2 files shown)


bmalehorn:
Neat commit. Certainly we all want more .gitignore to be more accurate.

But, this commit only targets a certain case, the wildcard ignore. It doesn't fix the underlying problem.

Say you have the file `file.txt`. This change will still fail to select `file.txt` with this `.gitignore`:

    file*
    !*.txt

Also, it will incorrect select file.txt with this `.gitignore`:

    *
    !*.txt
    *

Really the change should happen in `path_ignore_search()`, iterating through each regex and ignoring and un-ignoring the file as you go along (no "early exit" as you've programmed in).

But I just tried that and it failed because we sort `.gitignores` as part of an optimization. So the above `.gitignore` will be turned into

    !*.txt
    *
    *

which is *not* the same thing. So we'd have to throw out that sorting code, or modify it quite a bit to make sure it's still accurate in the face of un-ignores. Which is probably why un-ignores are not supported at the moment.

tl;dr to really fix the problem, rather than hack this one case, you need to do a lot of work. But you're welcome to give it a try :)

ssbarnea:
I cannot believe I spend almost an hour trying to find workarounds for this bug before finding a 3 year old PR that is fixing it. Is this project dead?

Issue #108, opened 2012-11-14
Line number is not printed in plugin 'Search in project' for Sublime Text 2


stigkj:
Not sure what happens here, but the line number is not part of the output from `ag` when Sublime Text 2's plugin 'Search in project' is running it.

The same happens when running it "manually" through Sublime Text's console, that is, by using Python's `subprocess.Popen(...)` in the console. The output from this is:

```
<filename>:<line>
```

If I try to include the `--column` option, this is the output:

```
<filename>:<column>:<line>
```

alphapapa:
#943 

It's been almost 6 years, I wish we could get this fixed...

PR #945, opened 2016-07-25
options.c: ignored pipe no longer blocks context

### LINKED ISSUE #943: Wrongly treat regular file as stream
It seems that `ag` treat all the files in the while loop below as stream, thus it won't print the matching line numbers nor matching filename (heading) after finish searching:

```bash
find . | while read -r f; do ag -H 'mySearchPattern' "$f"; done
```

I was trying to debug and looking at the code and found the offended line at

https://github.com/ggreer/the_silver_searcher/blob/52b8d5b361ebc160f97eb8cdd778c9ef46d9fcc3/src/options.c#L327

I don't have experience dealing with the file system so I'm not sure how to fix this. Really appreciate if anyone can explain or solve this issue.

Cheers.

bmalehorn:
Before:

```
$ true | ag -nH hello file.txt
hello
```

After:

```
$ true | ag -nH hello file.txt
file.txt
1:hello
```

If `ag` gets a pipe and files, it will only search the files.
But before, the presence of an ignored pipe would inhibit line number
and filename printing.

Fixes #943.

This PR also includes a test for this behavior, `ignore_pipe.t`, and fixes a test that was relying on this bug.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -810,6 +810,10 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
         }
     }
 
+    if (opts.passthrough) {
+        opts.print_line_numbers = 0;
+    }
+
 #ifdef _WIN32
     windows_use_ansi(opts.color_win_ansi);
 #endif
--- a/src/options.c
+++ b/src/options.c
@@ -739,14 +739,6 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
         }
     }
 
-    if (opts.search_stream) {
-        opts.print_break = 0;
-        opts.print_path = PATH_PRINT_NOTHING;
-        if (opts.print_line_numbers != 2) {
-            opts.print_line_numbers = 0;
-        }
-    }
-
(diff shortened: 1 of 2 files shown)


tests/passthrough.t
```diff
@@ -10,7 +10,7 @@ Setup:
 No impact on non-stream:
 
   $ ag --passthrough zoo passthrough_test.txt
-  zoo zar
+  2:zoo zar
```
ggreer: `--passthrough` is not supposed to add output. Adding a `2:` is incorrect behavior.

bmalehorn:
Thanks for the feedback. I've now fixed the issue with `--passthrough` properly. Thoughts?

outcoldman:
I believe this will fix #997 and #108, can this be merged?

bmalehorn:
@outcoldman I get the impression that `ag` isn't actively maintained anymore.

alphapapa:
@ggreer Could we please get this PR merged?  This bug has existed for over 5 years.

bmalehorn:
Closing due to inactivity.

PR #947, opened 2016-07-28
Added tpp as c++ extension

### LINKED ISSUE #946: Include .tpp as a C++ extension
Some C++ projects ([example](https://github.com/ComputationalRadiationPhysics/picongpu)) with very template-heavy code bases split their headers into "regular" `.hpp` files and `.tpp` files that are included like headers but contain the template implementations.

I suggest adding `.tpp` as one of the extensions that is seached when using the `--cpp` switch.

jschpp:
Fixes #946


ggreer:
C++ sure does have a lot of extensions. :)

Thanks for the PR.

PR #951, opened 2016-08-05
Fix command line formatting in markdown man page

fritzo:


ggreer:
This markdown file is a template to generate a manpage. It's not meant to be viewed by users directly.

PR #953, opened 2016-08-12
Fix typo: grep -> git

jonathn:


ggreer:
I totally forgot I even had a branch for the GitHub pages stuff. Thanks for the fix.

Issue #956, opened 2016-08-15
ag not printing empty files with -ul


HaleTom:
From the man page:  (`ag version 0.28.0`)

> ```
>    -l --files-with-matches
>           Only print the names of files containing matches, not the matching lines. An empty query will print all files that would be searched.
> 
>    -u --unrestricted
>           Search all files. This ignores .agignore, .gitignore, etc. It searches binary and hidden files as well.
> ```

So I'd expect `ag -lu` to list all files.

However:

```
+ravi@boxy:~$ mkdir new && cd new
+ravi@boxy:~/new$ echo stuff > non-empty && touch empty
+ravi@boxy:~/new$ ag -ul
non-empty
+ravi@boxy:~/new$
```

HaleTom:
There has been some discussion (including code) at:
[How do I force ag (the sliver searcher) to list empty files](http://unix.stackexchange.com/questions/303326/how-do-i-force-ag-the-sliver-searcher-to-list-empty-files)

ggreer:
Fixed in #957. Thanks!

PR #957, opened 2016-08-25
Print empty files with empty query and -ul

HaleTom:
Fix issue #956:

`ag -ul` with an empty query should list all filenames, as the manual
says:

  -u --unrestricted
      Search _all_ files. ...

  -l --files-with-matches
      ...An empty query will print all files that would be searched.

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -258,7 +258,11 @@ void search_file(const char *file_full_path) {
     f_len = statbuf.st_size;
 
     if (f_len == 0) {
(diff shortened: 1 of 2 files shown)


HaleTom:
Mentioned at http://unix.stackexchange.com/questions/303326/how-do-i-force-ag-the-silver-searcher-to-list-empty-files/322903#322903

ggreer:
Thanks.

PR #959, opened 2016-09-01
Output logging messages all in one shot

mernst:
ag uses multiple processes, so outputting messages piecemeal caused different messages to get interleaved.  That made them hard to use and interfered with downstream tools.

This commit reduces the likelihood of the problem (I used to see the interleaving many times per day), but I have never seen it in weeks of use with this patch), by constructing an entire line of output before passing the whole thing to vprintf.  The performance impact is negligible.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -45,24 +45,29 @@ void vplog(const unsigned int level, const char *fmt, va_list args) {
 
     FILE *stream = out_fd;
 
+    char *level_fmt = malloc(strlen(fmt) + 9);
+    level_fmt[0] = 0;
     switch (level) {
         case LOG_LEVEL_DEBUG:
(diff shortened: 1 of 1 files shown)


ggreer:
This has race conditions. For example, when running `ggreer@zinc:~/code% ./ag/ag --workers=8 --debug blah`, I'll get occasional lines like this:

    DEBUG: DEBUG: %s added to work queue
     added to work queue

The real fix is to lock/unlock the print mutex. I'll create a PR pretty soon.

mernst:
Agreed.  I mentioned that my approach "reduces the likelihood" (and made ag usable rather than unusable for me), but your solution is better.  Thanks for the fix!

PR #960, opened 2016-09-02
Use the print mutex when logging.

ggreer:
Fixes occasional glitches in log messages. Related to #959.


ggreer:
Note that this does hurt performance when running ag with --debug. There's really no getting around this.

PR #962, opened 2016-09-06
Replaced the links to ag.vim with the active fork.

manasthakur:
`rking/ag.vim` has been deprecated; the active fork is maintained by @albfan at `albfan/ag.vim`.


albfan:
Good to hear, but I'm in doubt that's a fact. All my work is related to `ag --group` usage. Do as you will, I use it on regular basis and issues or patches are welcome.

PR #971, opened 2016-09-21
add zsh completion function

okapia:
It would be great if you could include this completion function for zsh. There's already one for bash. I've also patched the Makefile to install it into $(datadir)/zsh/site-functions which is typically a directory in which zsh will find it provided that it is named with an initial underscore. Thanks


okapia:
I've updated the commit to cover changes following the 0.33 release

ggreer:
Thanks!

PR #974, opened 2016-09-24
Prefer .ignore to .agignore.

ggreer:
Thanks to discussion on https://news.ycombinator.com/item?id=12564442, @BurntSushi and I agreed on a common file name for ignore patterns. [He's made his PR for ripgrep]( https://github.com/BurntSushi/ripgrep/issues/40). This is ag's counterpart.

To those who worry about change: Please don't! It will be a long time before .agignore is no longer obeyed, if ever.

### DIFF
--- a/doc/ag.1.md
+++ b/doc/ag.1.md
@@ -153,12 +153,12 @@ Recursively search for PATTERN in PATH. Like grep or ack, but faster.
     Search all text files. This doesn't include hidden files.
 
   * `-u --unrestricted`:
-    Search *all* files. This ignores .agignore, .gitignore, etc. It searches
+    Search *all* files. This ignores .ignore, .gitignore, etc. It searches
     binary and hidden files as well.
 
   * `-U --skip-vcs-ignores`:
     Ignore VCS ignore files (.gitignore, .hgignore, svn:ignore), but still
-    use .agignore.
+    use .ignore.
 
   * `-v --invert-match`:
     Match every line *not* containing the specified pattern.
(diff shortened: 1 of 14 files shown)


doc/ag.1.md
```diff
@@ -195,13 +195,13 @@ run `ag --list-file-types`.
 ## IGNORING FILES
 
 By default, ag will ignore files whose names match patterns in .gitignore,
-.hgignore, or .agignore. These files can be anywhere in the directories being
+.hgignore, or .ignore. These files can be anywhere in the directories being
 searched. Ag also ignores files matched by the svn:ignore property if `svn
 --version` is 1.6 or older.  Finally, ag looks in $HOME/.agignore for ignore
```
jonathanjouty: > Finally, ag looks in $HOME/.agignore

Shouldn't it now say `$HOME/.ignore`?

jonathanjouty:
Good work on agreeing to commonly use `.ignore`! Only just caught up with this on HackerNews, so sorry for being late to the game...

I have a minor quibble...
As you're keeping support for `.agignore`, you should mention in the documentation that this *is still* part of current behaviour, and will be discontinued on some future release.

I imagine some poor soul `git clone`ing a repository with a `.agignore`, breaking their head over *why* they can't `ag` what they are looking for...

Also relevant for https://github.com/BurntSushi/ripgrep/pull/41

davidosomething:
In case anyone needs clarification:
- `CURRENTDIR or PARENTDIRS/.ignore` is the local ignore file
- `~/.agignore` is global and included in every search, pretty much appended to the bottom of your local `.ignore`

rafi:
It would be great if ag could search for it in `~/.config/ag/ignore` also.

doc/ag.1.md
klas-genestack: Seems like *ag 1.0.3* doesn't take `~/.ignore` into account at all (still ignoring patterns from `~/.agignore`).

krigstask: Seems like *ag 1.0.3* doesn't take `~/.ignore` into account at all (still ignoring patterns from `~/.agignore`).

mMontu: A little late, but it may be useful for someone else until #1084 is done: according to the [wiki](https://github.com/ggreer/the_silver_searcher/wiki/Advanced-Usage)
> consider adding this alias: `alias ag='ag --path-to-ignore ~/.ignore'`

PR #979, opened 2016-09-29
escape usage in markdown

jbodah:
Usage doesn't render properly on Github unless we escape the `<>`


ggreer:
Thanks for noticing this. I fixed it in 052cb33cdc02235197e3c791a4be63a6c13f0a2b, and I made sure to give you credit.

Issue #980, opened 2016-10-02
Feature request: Support both --no<option> and --no-<option>


okdana:
A while ago you decided to standardise the option negation prefix to `no` without a hyphen — so `--nogroup` as opposed to `--no-group`. Whilst standardising them is great, i find that, even with using `ag` almost every day, i *still* get confused about it.

Much of that seems to be habit instilled by other software i've used. I checked the man pages of my top 20 most used commands, and almost every single one that had a 'no'-style option used `no-` with the hyphen — `git`, GNU `grep`, `rsync`, `psql`, `php`, `composer`, `brew`, GNU `make`, &c. The only one i could find that used `no` without the hyphen was `vim`. And, personally, i think it only *makes sense* to use the hyphen when you're using it as a word separator in options elsewhere (as `ag` does with `--case-sensitive`, `--max-count`, &c.).

With that said, the obvious solution to please everyone and maintain backwards compatibility would be to just support both forms. They don't both have to be documented or shown in the usage help, just silently accept them and everybody's cool.

Would that be OK with you? If so, i can put together a PR. Let me know how you feel. Thanks again!

ggreer:
I chose no hyphen because that's what ack v1 did. Now that ack v2 is out and its features have diverged from ag, I guess it doesn't make much sense. You also make some good points about other tools and how every other word is separated by a dash.

If you want to make a PR, go for it.

okdana:
Cool! Closing in favour of PR #984.

PR #984, opened 2016-10-07
Support both --no-<option> and --no<option> (addresses #980)

okdana:
This PR addresses feature request #980.

ETA: Added some annotations to the diff, please review

ETA2: I left all of the documentation as it was (they all say `--[no]option`) — let me know if you want me to change that though

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -260,20 +260,30 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
         { "max-count", required_argument, NULL, 'm' },
         { "mmap", no_argument, &opts.mmap, TRUE },
         { "multiline", no_argument, &opts.multiline, TRUE },
-        /* "no-" is deprecated. Remove these eventually. */
-        { "no-numbers", no_argument, &opts.print_line_numbers, FALSE },
-        { "no-recurse", no_argument, NULL, 'n' },
+        /* Accept both --no-* and --no* forms for convenience/BC */
+        { "no-affinity", no_argument, &opts.use_thread_affinity, 0 },
         { "noaffinity", no_argument, &opts.use_thread_affinity, 0 },
+        { "no-break", no_argument, &opts.print_break, 0 },
         { "nobreak", no_argument, &opts.print_break, 0 },
+        { "no-color", no_argument, &opts.color, 0 },
         { "nocolor", no_argument, &opts.color, 0 },
+        { "no-filename", no_argument, NULL, 0 },
         { "nofilename", no_argument, NULL, 0 },
+        { "no-follow", no_argument, &opts.follow_symlinks, 0 },
         { "nofollow", no_argument, &opts.follow_symlinks, 0 },
+        { "no-group", no_argument, &group, 0 },
         { "nogroup", no_argument, &group, 0 },
+        { "no-heading", no_argument, &opts.print_path, PATH_PRINT_EACH_LINE },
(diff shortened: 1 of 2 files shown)


src/options.c
```diff
@@ -260,20 +260,30 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
         { "max-count", required_argument, NULL, 'm' },
         { "mmap", no_argument, &opts.mmap, TRUE },
         { "multiline", no_argument, &opts.multiline, TRUE },
-        /* "no-" is deprecated. Remove these eventually. */
-        { "no-numbers", no_argument, &opts.print_line_numbers, FALSE },
-        { "no-recurse", no_argument, NULL, 'n' },
+        /* Accept both --no-* and --no* forms for convenience/BC */
+        { "no-affinity", no_argument, &opts.use_thread_affinity, 0 },
```
okdana: I couldn't find a really elegant way for `getopt` to handle hyphenated and non-hyphenated variants, so i pretty much just copied and pasted everything in this block of options 💁

src/options.c
```diff
@@ -505,10 +515,18 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
                 } else if (strcmp(longopts[opt_index].name, "ignore") == 0) {
                     add_ignore_pattern(root_ignores, optarg);
                     break;
+                } else if (strcmp(longopts[opt_index].name, "no-filename") == 0) {
```
okdana: This block of logic down here could possibly (?) be improved by assigning a `val` to the `option` structs for each of the short-option-less long options, and then handling them as a `case` in the main `switch` body.

For example:

```
enum {
	OPTION_NO_FILENAME = 1000,
	// ...
};

// ...
{ "no-filename", no_argument, NULL, OPTION_NO_FILENAME },
{ "nofilename", no_argument, NULL, OPTION_NO_FILENAME },
// ...

switch (ch) {
	// ...
	case OPTION_NO_FILENAME:
	  opts.print_path = PATH_PRINT_NOTHING;
	  opts.print_line_numbers = FALSE;
	  break;
	// ...
}
```

Since it was only `--no-filename` and `--no-pager` that were really affected, though, i just left it alone. Let me know if you'd prefer something else.

src/options.c
ggreer: There is no elegant way to do this. Welcome to ancient C APIs! 🎉

src/options.c
ggreer: Instead of copy-pasting the body of the conditional, I recommend something like:

```
} else if(strcmp(longopts[opt_index].name, "nofilename") == 0 ||
          strcmp(longopts[opt_index].name, "no-filename") == 0) {
```

Ditto for `no-pager`.

okdana: Oh, yeah. I looked around to see how you would have handled a long condition like that, but i couldn't find any examples in the source, so i just gave up 🙍 But yeah i'll change that

okdana: Addressed this feedback. Let me know if you need me to squash the commits before you merge 🎃

ggreer:
Thanks for making this PR. I'd like you to fix the copy-pasted code I mentioned in https://github.com/ggreer/the_silver_searcher/pull/984#discussion_r82495967. Besides that, it looks good to me. If for some reason you take a while to respond, I'll merge your PR and fix that myself.

ggreer:
🎉

Issue #999, opened 2016-10-19
add .log filetype restriction


ently:
Love ag, it'd be great if log files could be added to the list of filetype restrictions

PR #1002, opened 2016-10-29
Add support for log file type.

spencercrissman:
Implements #999.


tests/list_file_types.t
```diff
@@ -131,6 +131,9 @@ Language types are output:
   
     --lisp
         .lisp  .lsp
+
```
jschpp: There is too much space here.

spencercrissman:
Sorry about that. I've adjusted the whitespace to match the others.

PR #1003, opened 2016-11-01
Add PureScript support

sotojuan:


sotojuan:
Any way to know what went wrong?

jschpp:
You can have a look at details at travis.
But the problem here is incorrect whitespaces. Compare the whitespaces in your changes to the lines around them.

CarlosMChica:
Is there a reason why this didn't get included? It seems it was just because the build was failing, in which case I'd be happy to create a new PR myself.

Please let me know.
Thanks.

PR #1015, opened 2016-11-22
Minimize malloc calls in ag_scandir

forgottenswitch:
Previously, `malloc(3)` was called for every entry in a dir,
results were accumulated, and freed in a loop after a dir was done.

Implement dedicated array to minimize malloc calls.
Each dir now only does one malloc, one free, and several realloc's.

This should have better cache density.
At least, it cannot be slower.

Inspired-by: https://speakerdeck.com/alex/why-python-ruby-and-javascript-are-slow

### DIFF
--- a/src/scandir.c
+++ b/src/scandir.c
@@ -10,9 +10,14 @@ int ag_scandir(const char *dirname,
                void *baton) {
     DIR *dirp = NULL;
     struct dirent **names = NULL;
+    struct dirent *ents = NULL;
+
+    size_t names_len = 32;
+    size_t ents_bsz = 4000, ents_blen = 0;
+
     struct dirent *entry, *d;
-    int names_len = 32;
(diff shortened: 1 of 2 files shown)


forgottenswitch:
This also fixes a voodoo which mystically never failed:
```
#if defined(__MINGW32__) || defined(__CYGWIN__)
        d = malloc(sizeof(struct dirent));
#else
        d = malloc(entry->d_reclen);
#endif
```
How did it manage to hold `sizeof(struct dirent) + d_reclen` in latter, when first is 280 and latter is ~20 ?

forgottenswitch:
This part somehow wasn't crashing; how?
```
#if defined(__MINGW32__) || defined(__CYGWIN__)
        d = malloc(sizeof(struct dirent));
#else
        d = malloc(entry->d_reclen);
#endif
```

Issue #1016, opened 2016-11-22
Filetype option for listing directories like `find -type d`?


kevinzwhuang:
I noticed there isn't a directory/folder type in --list-file-types. I want to list directories like you would with `find -type d`, is this possible with ag?

forgottenswitch:
Would it be of any use?

kevinzwhuang:
Definitely @forgottenswitch. One great use case would be to allow ag to replace the default ALT-C command for FZF (which uses `find`) to quickly search and switch directories with `cd`.

See https://github.com/junegunn/fzf for more info - replacing the main command CTRL-T is already doable with `export FZF_DEFAULT_COMMAND='ag -g ""'`

forgottenswitch:
So you want something like `find | ag ...` ?

forgottenswitch:
```
#!/bin/sh

files=$(ag -g "$@")
dirs=$(echo "$files" | sed -e 's:/[^/]*$::' | uniq)

echo "$files"
echo "$dirs"
```

forgottenswitch:
And why `fzf` cannot do the above on its input?

kevinzwhuang:
fzf already can - it uses `find` for it by default, something like `find -type d`.

That workaround might end up being too slow and would ignore empty folders.

I'm asking to see if it would be possible to do the same with ag, though it seems maybe it won't be possible since it does not print directories.

forgottenswitch:
But the script above does just that.

forgottenswitch:
Not that it is not possible, but matching on directory names is something ag wasn't meant for :)

forgottenswitch:
See https://github.com/forgottenswitch/the_silver_searcher/tree/gdirs
Reopening the pull once it is OK.

forgottenswitch:
Go has `filepath.Walk`; use that in either fzf or helper binary.

forgottenswitch:
To rephrase the request:
- `--dirs` for printing directories with `-g`
- `--dirs-only` for not printing files with `-g`
- `--dirs-nonempty` for ignoring empty directories with `-g`
And if *that* is implemented, then why not:
- `--dirs-with-files-matching`
- `--dirs-with-filenames-matching`
- `--dirs-with-dirnames-matching`

forgottenswitch:
Maybe `--dirs`, `--dirs=nonempty`, and `--dirs=only` shold be added.

acannon828:
+1

[...]

PR #1018, opened 2016-11-23
Add -g PATTERN --print-dirs

forgottenswitch:
Functionality for https://github.com/ggreer/the_silver_searcher/issues/1016
Maybe there should also be `--print-dirs-only`.


theonlygusti:
When will `--print-dirs` be added?

PR #1019, opened 2016-11-23
Simpler . and .. filtering

forgottenswitch:
Previously, "." and ".." were looked up using strcmp() in
a builtin-ignores list.

Compare them manually char-by-char instead.
This should avoid unneeded calls, returns, array traversing,
and string dereferencing.

Hardcoding .pyc, etc. by default hardly is worth it,
as without them in .gitignore already, user experience
is improper anyway.

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -20,14 +20,6 @@
 const int fnmatch_flags = FNM_PATHNAME;
 #endif
 
-/* TODO: build a huge-ass list of files we want to ignore by default (build cache stuff, pyc files, etc) */
-
-const char *evil_hardcoded_ignore_files[] = {
-    ".",
-    "..",
-    NULL
-};
-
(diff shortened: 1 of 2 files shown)


ggreer:
Did you actually benchmark/profile this? It seems to have no difference in performance for me.

forgottenswitch:
No I did not.

> Hardcoding .pyc, etc. by default hardly is worth it,
as without them in .gitignore already, user experience
is improper anyway.

forgottenswitch:
@ggreer the idea was to move attention from built-in rules to sharing-with-git-or-anything-else.
I.e. force (in good faith) the user to maintain proper `.gitignore` and `~/.config/git/ignore`.
Accordingly, do not confuse ag source reader with builtin-ignores idea.

forgottenswitch:
"Micro-optimization" was the term.

PR #1028, opened 2016-12-01
Fix broken link of ack.vim

keiko713:
Seems like the link to ack.vim was dropped at https://github.com/ggreer/the_silver_searcher/commit/787d9e3261b0b4b07a31d6d8f7dbf67713b1ddea , so putting it back.


jschpp:
@ggreer why close this. you have a broken link in line 144 of README.md

PR #1032, opened 2016-12-11
Create autogen.sh script

aswild:
Many autotools-based projects have an autogen.sh script which handles
creating the configure script properly. The existing build.sh does this
but doesn't offer a way to just run autotools without configure/make.

Here I split the first part of build.sh into autogen.sh and then call
autogen.sh from build.sh. Using "set -e" cleans up the script a bit to
avoid a long chain of "&&"s

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+set -e
+cd "$(dirname "$0")"
+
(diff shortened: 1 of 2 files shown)


aswild:
Thanks for the merge @ggreer! Considering how many PRs were open here I figured it'd end up sitting for a while. :)

Issue #975, opened 2016-09-28
Cannot searc files > 2Gb


ggl:
If I try to search a file bigger than 2Gb, I get the follwing error:
ERR: Skipping system.log: pcre_exec() can't handle files larger than 2147483647 bytes

Grep and ack both work fine (although for ack it takes like forever).

jschpp:
This is by design. The regex Engine (PCRE) can't handle files that large.
You can find [here](http://pcre.org/original/doc/html/pcrelimits.html) that the maximum subject (file) length is `INT_MAX` which is 2147483647 for a signend 32bit int. Therefore the maximim file size is `INT_MAX` in bytes.

ggl:
Most uses of grep/ack/ag are line by line searches. You would only have to reach the maximum subject length on multiline searches or if a single line is over 2G. So for most other uses ag would only need to match a single line at a time.

jschpp:
You are right in saying most searches are single line only. Nevertheless ag does multi-line searching by default (as far as I know it matches newlines with the `\s` regex). 
I found that files greater than 2GB can be searched with a literal (not regex) pattern. In theory ag could make a case-by-case decision and only raise that error in case of a single line greater `INT_MAX` bytes or multiline searching. 

Maybe @ggreer could mention whether he wants this or not. I'm not sure how much work it would be to patch this to support the above case-by-case choice.

netheril96:
pcre has a new version called pcre2 with backwards incompatible new API. The new API uses `size_t` instead of `int` to refer to lengths, so it can handle strings larger than 2GB. Maybe `ag` should update to require pcre2 instead.

njt1982:
@jschpp how do you do literal patterns?

jschpp:
@njt1982 From `ag --help`: `  -Q --literal            Don't parse PATTERN as a regular expression`

dhivyada:
Any updates on upgrade to pcre2?

monperrus:
I'm hit by this limitation today to grep my 7GB mailbox. It would be really great to handle large files.

jeffythedragonslayer:
I hit these errors today when running ag from my home directory and then ag dumped core.  Too bad I didn't have ulimit -c unlimited enabled

dimaqq:
Is there maybe a change for a command line argument that would mean something like "process first 2GB of data, then give up"?

pdelteil:
You could add a flag to split the file in 2GB parts and then merge the result of every run.

PR #1035, opened 2016-12-18
Add support for libpcre2 (compiles and works)

aswild:
I did a thing.

Add optional support to build with libpcre2.
Enable by passing --with-pcre2 to ./configure, the default is still to
build against legacy libpcre.

To facilitate compiling against either PCRE API version, an ag_pcre
wrapper API was created. This is done through a combination of AG_PCRE_*
defines and ag_pcre_* functions, plus typedefs ag_pcre_re_t and
ag_pcre_extra_t

At this point, ag can be built with libpcre or libpcre2 and passes tests
without modifying them. Preliminary testing hasn't shown much change in
speed, but further profiling should still be done.


gcflymoto:
this is great, I cherry picked your changes and updates for the latest code in git. Hopefully with this patch it solves https://github.com/ggreer/the_silver_searcher/issues/975

aswild:
out of date PR which got no attention anyway

PR #1036, opened 2016-12-19
tests: Sort the files list in empty_match.t to fix ordering issues (fixes #1029)

### LINKED ISSUE #1029: tests/empty_match.t: last test fails due to file ordering issues
The empty.txt and nonempty.txt files can be returned in either order causing the test to fail some of the time. Sorting the output in some fashion is probably required, but a simple use of sort can't be used due to the stats also being included in the output.

aswild:
Issue #1029 - fixes intermittent failures due to the order of file searching

### DIFF
--- a/tests/empty_match.t
+++ b/tests/empty_match.t
@@ -13,10 +13,17 @@ A genuine zero-length match should succeed:
   1:foo
 
 Empty files should be listed with --unrestricted --files-with-matches (-ul)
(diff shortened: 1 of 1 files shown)


ggreer:
How's https://github.com/ggreer/the_silver_searcher/commit/5444df961283d288eb20f1dbaf3ad6661a53b8c3 instead?

aswild:
Yeah, that looks like a simpler/cleaner solution.

PR #1037, opened 2016-12-19
Add support for ~/.agrc

aswild:
This has been done before by a variety of users, but most of the PRs to add this feature have been dormant for months or years, so here's my implementation.

------
Using shell aliases for ag to change the defaults (such as --color-match)
is ugly and breaks trying to use ag nicely in cases like xargs.
As such, it's very helpful to have a default ".agrc" file which can be
read to apply custom options.

The .agrc file file should contain a list of command-line options, one
per line. Only lines which start with '-' are used, all others are
ignored.

The default agrc file is $HOME/.agrc, but can be overridden with
--agrc=<agrc-file>
--no-agrc or --noagrc are supported to disable reading from agrc.
There is no support yet for multiple agrc files.

Tests have also been updated to pass --noagrc to keep a user's defaults
from failing tests.

### DIFF
--- a/doc/ag.1
+++ b/doc/ag.1
@@ -23,6 +23,10 @@ Output results in a format parseable by AckMate \fIhttps://github\.com/protocool
 Set thread affinity (if platform supports it)\. Default is true\.
 .
 .TP
+\fB\-\-agrc=<path\-to\-agrc\-file>\fR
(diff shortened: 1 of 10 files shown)


aswild:
I've identified a couple memory leaks (fixed mostly in my master branch) and need to update this PR branch

PR #1049, opened 2017-01-25
Fix cleanup when mmap returns MAP_FAILED

lpenz:
MAP_FAILED != NULL, which means that, when that error is returned,
the code was trying to munmap it.

We now set buf = NULL before going to cleanup.


ggreer:
Instead of setting buf to `NULL`, causing a no-op side-effect of `free(NULL)`, [I added a check for `MAP_FAILED`](https://github.com/ggreer/the_silver_searcher/commit/7110453f8c7e844f2856073dbfd2ee935f43e1a8) and gave you credit in the commit message. Is that OK?

lpenz:
That's perfect, thanks!

PR #1059, opened 2017-02-15
Support for gradle files

realdeadfish:
Add `.gradle` files to the list of known files to search for when filtering for `--groovy`.


realdeadfish:
Hrm... I'd luckily execute a build (and test) locally, but `./configure` gives me

```
./configure: line 3954: syntax error near unexpected token `PCRE,'
./configure: line 3954: `PKG_CHECK_MODULES(PCRE, libpcre)'
```

Now, my autoconf foo is a bit old, all I did was `autoreconf -i && ./configure`. autoreconf 2.69 / automake 1.15 from Homebrew / macOS.

PR #1061, opened 2017-02-23
Add "twig" to languages

jdeniau:
Here is a proposal to add `twig` files, a template file for [twig](http://twig.sensiolabs.org/).

By the way, Isn't it a little "overkill" to add this in the language file,  because it's a very particular file type ?

I did not found a proper way to declare a custom file type just for my usage.

### DIFF
--- a/src/lang.c
+++ b/src/lang.c
@@ -95,6 +95,7 @@ lang_spec_t langs[] = {
     { "tt", { "tt", "tt2", "ttml" } },
     { "toml", { "toml" } },
     { "ts", { "ts", "tsx" } },
+    { "twig", { "twig" } },
     { "vala", { "vala", "vapi" } },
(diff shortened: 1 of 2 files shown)


jschpp:
The language file is only used when searching for a specific language/extension. So it's not that much of an overkill.
Your tests are failing because you are missing a test case for twig files.
You need to add a test case to [tests/list_file_types.t](https://github.com/ggreer/the_silver_searcher/blob/master/tests/list_file_types.t)
```diff
--- a/tests/list_file_types.t
+++ b/tests/list_file_types.t
@@ -276,6 +276,9 @@
     --ts
         .ts  .tsx
   
+    --twig
+        .twig
+  
     --vala
         .vala  .vapi
   
```
Please mind the whitespaces. The tests will fail if they are missing.

jdeniau:
I didn't saw this either.

I updated my PR, feel free to close it if you prefer the `templates` approach

Issue #1027, opened 2016-11-29
How to search in UTF8 documents?


monperrus:
I'd like "montr.al" to match both "montreal" and "montréal". When the latter is encoded in UTF8, the `é` is coded on two bytes and is not matched by default with "montr.al" by ag.

How to do this?

mensfeld:
@monperrus  did you find solution for that one?

monperrus:
nop.

mensfeld:
@monperrus  if you are still interested in that, blog.burntsushi.net/ripgrep/ works with utf-8

JFLarvoire:
I've fixed this in the Windows build that I'm working on.  
Sources in https://github.com/JFLarvoire/the_silver_searcher  
Prebuilt versions in http://jf.larvoire.free.fr/temp/ag.zip  

For the Unix version, as a workaround, you can enable the UTF8 regexp search mode by prefixing the regexp with (*UTF8). This is a little awkward, but it should work if PCRE itself has been compiled with UTF8 capabilities:  
'ag (*UTF8)montréal'

Eventually this should be automated in the Unix version too if the locale is a UTF8 locale.

PR #1064, opened 2017-03-05
Support for building a Windows version with Microsoft Visual C++

JFLarvoire:
Hello,

This version builds ag for Windows with Microsoft Visual C++.  
It is loosely based on the version that KJK did a few years ago, and that is mentioned in your README.md.

Mine has the following changes compared to his:

- The ag sources, and the included libraries sources, are updated to the latest versions.  
   => Minimal changes to the ag src directory.
- The build procedure does not require any external development tool (besides Visual C++), and it's extremely simple.
- It should work with all versions of Visual C++ from VC++ 2010 to VC++ 2015. (VC++ 2017 is coming soon, and I'll add support asap.) (Older versions do not support C99, which is required by ag own sources.)
- It uses the MsvcLibX library to add the missing routines in Microsoft libraries, which brings many significant advantages compared to KJK's version:
- It uses UTF-8 internally, and outputs file paths correctly in any Windows code page.
  (This is very important to me, as my first name contains a non-ASCII character, and KJK's version did not display my HOME path correctly!)
- It fixes a number of bugs, including the 260-character limit in the basic Microsoft C libraries.
  (I do have systems with some paths longer than that!)
- Last but not least, it still builds successfully in Unix.

I know this looks like a huge merge, but the changes to the existing ag sources are actually very minor.
Most of the files are new files in the msc_libs subdirectory, and are copied from other projects.  
Note that all these other project have compatible licenses: Either Apache 2 like ag itself; Or BSD for some; Or LGPL for others.

Side note: I'm using PCRE regular expressions in UTF-8 mode, which fixes issue #1027 for Windows.  
I can work with you to fix it for Unix too.

There are explanations in the updated README.md, and more detailed ones in the msc_libs/README.md.

Before we go any further, maybe it'd be good to have the build process tested by other developers.
Is there a distribution list, where I post a message asking for volunteers?

Best regards,

Jean-François Larvoire


JFLarvoire:
I tried running './format.sh reformat' to fix formatting, as suggested in the detailed reports.  
But this makes lots of changes to the existing files, in places that I've NOT changed.  
Just in case, I've checked in a minor change, removing a change that causes a warning in the report, and that I did not really need. (I did it while investigating the pthread 2.10 RC issue, and forgot to remove it.)  
What should I do now to fix these checks?

jschpp:
I added a PR to your repo with clang-format 3.8 JFLarvoire/the_silver_searcher#2

jschpp:
Nicely done. And the tests run without errors 👍 
Although there are still some compiler warnings

JFLarvoire:
The problem is not with my code anymore, it's with the checking process:
`./format.sh reformat` simply breaks the Windows batch file and the two make files that I've hade to add into the src directory.
How can I fix format.sh so that it checks only *.c and *.h files?

JFLarvoire:
Unrelated: I've had a nasty issue with threading in MsvcLibX, which forced me to make a change in ag's log_debug routine: 
log_bug now displays the worker thread ID ahead of the debug string. This allows filtering the debug output by thread number, to make sense of who's doing what.
I've left this in, as the change is relatively unobtrusive, and I feel this might be useful to others.
The issue in MsvcLibX is now fixed :-) (Some code used a static variable, which was not thread safe.)

JFLarvoire:
I finally modified format.sh to only reformat src/*.c and src/*.h files, and the test now passes :-)

JFLarvoire:
I'll need help setting up clang-format:

I've uploaded a couple of days ago a set of sources that work... But "make test" fails on github because of clang-format errors.
I've struggled the whole day yesterday trying to make it work on my PC, but no matter what I do, this fails:

- In Windows subsystem for Linux, which is a plain Ubuntu, the ag build works,
  but clang-format segfaults after modifying a couple of files.
- I tried installing clang-format in a CentOS VM, to rebuild everything in a true Linux environment...
  Again, clang-format segfaults after modifying files differently!
- Finally I realized that a Windows version of clang-format does exist. I installed it.
  This one successfully processes all files (Bypassing format.sh, and running clang-format directly with the same options), but with still different results!

And in ALL 3 CASES, when I push the modified files I have to github, the version of clang-format there rejects them as wrongly formatted!!!

Could anybody explain me how to setup a development environment with clang-format, that works like the one used by the github build system?

Jean-François

osimola:
I had the same problem with clang-format. Different versions just format differently and there is no knowing which one Travis is running.

I solved the problem by copying the suggested diff from error report and applying it to my copy.

JFLarvoire:
Thanks @osimola, all checks have passed at last :-)

PR #1068, opened 2017-03-20
Added long option for filename pattern

### LINKED ISSUE #1030: Add long flag for -g
If you use `ag` in a script it's hard to make them easily understandable which such obscure flags like `-g`:

```shell
ag -g json
```

If you are not familiar with `ag` this could mean anything (search globally, group results, …),
but it's not obvious at all that it means *Print filenames matching 'json'*

We should add a long flag so it can be used to increase readability.
Possible suggestions:

```shell
--filename-pattern
--path-pattern # as it actually matches the whole path and not just the filename
--pathname
```

`--filename` is not an option as it is already taken (although the current one should rather be `--print-filename`, but I guess it's too late to fix this now…)

Also i don't think the search pattern should be an argument of the flag.
Rather should the normal pattern position as in `ag [FILE-TYPE] [OPTIONS] PATTERN [PATH]` be used.
This would mean that `ag --silent -g json` and `ag -g --silent json` were the same thing and that a long flag was easier understandable (i.e. the `-pattern` part wouldn't be necessary anymore).

jschpp:
Fixes #1030

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -57,7 +57,8 @@ Output Options:\n\
                           (Enabled by default)\n\
   -C --context [LINES]    Print lines before and after matches (Default: 2)\n\
      --[no]group          Same as --[no]break --[no]heading\n\
-  -g PATTERN              Print filenames matching PATTERN\n\
+  -g --filename-pattern PATTERN\n\
+                          Print filenames matching PATTERN\n\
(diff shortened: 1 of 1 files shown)


ggreer:
❤️

PR #1069, opened 2017-03-30
Boyer moore algorithm ignore case refactor

PRDeving:
I unifyed the boyer-moore algorithm to get another parameter as ignore-case flag, now the code is clearer and there's no need of func proxy anymore


ggreer:
You had a bug in your order of operations which I fixed in https://github.com/ggreer/the_silver_searcher/commit/04a373bafafd7fb9e5977af5e6fe00dd79764fc5

Thanks for cleaning up my code.

PRDeving:
Oh...that's why, sorry.

np, thanks you for this marvelous tool :D

PR #1083, opened 2017-05-09
add support for jsp tag and fragment files

ackl:
[Some java EE projects use .jspf, .tag and .tagf files.](http://docs.oracle.com/javaee/5/tutorial/doc/bnama.html)


ggreer:
Thanks!

PR #1091, opened 2017-05-27
Add --print-all-files option

uhlissuh:
### Motivation

We'd like to explore using `ag` for Atom's [project search](https://github.com/atom/find-and-replace). But in the project search UI, while the search is in progress, we need to display to the user the number of paths that have been searched so far. None of `ag's` current options would allow us to do that. The `--stats` flag shows the total number of files `ag` searched, but only displays this information *after* the search is complete, whereas we need to update the count in real time while `ag` is still searching.

### Solution

We thought it would make sense to add an option called `--all-files` to include a 'heading' for *every* file that `ag` searches, not just those that contain matches.

For example:

```sh
$ printf 'foo\n' > ./foo.txt
$ printf 'bar\n' > ./bar.txt
$ printf 'baz\n' > ./baz.txt

# Show a heading for `bar.txt` and `baz.txt`, even though they don't contain matches.
$ ag --all-files --group foo
bar.txt

baz.txt
  
foo.txt
1:foo
```

If you have any suggestions about the name of the flag or an alternative solution that would let us know the number of paths searched so far, we'd be glad to make the changes.

Paired with @maxbrunsfeld

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -63,6 +63,8 @@ Output Options:\n\
                           (don't print the matching lines)\n\
   -L --files-without-matches\n\
                           Only print filenames that don't contain matches\n\
+     --all-files          Print headings for all files searched, even those that\n\
+                          don't contain matches\n\
      --[no]numbers        Print line numbers. Default is to omit line numbers\n\
                           when searching streams\n\
   -o --only-matching      Prints only the matching part of the lines\n\
@@ -158,6 +160,7 @@ void init_options(void) {
     opts.path_sep = '\n';
     opts.print_break = TRUE;
     opts.print_path = PATH_PRINT_DEFAULT;
+    opts.print_all_paths = FALSE;
     opts.print_line_numbers = TRUE;
     opts.recurse_dirs = TRUE;
     opts.color_path = ag_strdup(color_path);
@@ -231,6 +234,7 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
         { "ackmate-dir-filter", required_argument, NULL, 0 },
         { "affinity", no_argument, &opts.use_thread_affinity, 1 },
         { "after", optional_argument, NULL, 'A' },
+        { "all-files", no_argument, NULL, 0 },
         { "all-text", no_argument, NULL, 't' },
         { "all-types", no_argument, NULL, 'a' },
         { "before", optional_argument, NULL, 'B' },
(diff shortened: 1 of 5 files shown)


src/options.c
```diff
@@ -231,6 +234,7 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
         { "ackmate-dir-filter", required_argument, NULL, 0 },
         { "affinity", no_argument, &opts.use_thread_affinity, 1 },
         { "after", optional_argument, NULL, 'A' },
+        { "all-files", no_argument, NULL, 0 },
```
ggreer: I hate to nitpick, but can the option be `--print-all-filenames` or something along those lines? `--all-files` makes me think "search all files" (like `-u`'s behavior).

src/options.c
```diff
@@ -510,7 +514,10 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
                 opts.path_sep = '\0';
                 break;
             case 0: /* Long option */
-                if (strcmp(longopts[opt_index].name, "ackmate-dir-filter") == 0) {
+                if (strcmp(longopts[opt_index].name, "all-files") == 0) {
+                    opts.print_all_paths = TRUE;
+                    break;
+                } else if (strcmp(longopts[opt_index].name, "ackmate-dir-filter") == 0) {
```
ggreer: I tend to keep these options alphabetized, so "ackmate-dir-filter" would go before "all-files".

ggreer:
I'd love for ag to be used by Atom. Change the option name and then this PR is good.

Also: now Atom devs are making a pull request to one of my projects? [How the tables have turned!](https://github.com/atom/atom/pull/11552) 😏

maxbrunsfeld:
Hey @ggreer, thanks for the review! I agree: `--print-all-filenames` is a better name. I paused work on this because we're still evaluating whether or not we can use `ag` in Atom. Since this option is only useful for a very specific situation (like Atom's), I would feel bad if you added it just for us unless we were definitely going to use `ag`.

The main thing we're concerned about is the handling of `.gitignore` files, e.g. https://github.com/ggreer/the_silver_searcher/issues/915#issuecomment-225764458. We're concerned that a lot of Atom users might hit this problem.

Thoughts? If you're cool with adding this option despite our uncertainty, I can definitely make the necessary adjustments. Otherwise, maybe we should close this until we decide whether Atom will be able to use it.

ggreer:
I added support for ! rules in ignores: #1101.

uhlissuh:
Wow, it's great to see that issue fixed! We're going to proceed assuming we can use `ag` in Atom, and we'll let you know if we find any other issues that might block us. We did the rename that you suggested.

ggreer:
Sounds good. I've tagged a new release (v2.0.0) that contains your PR (among other improvements).

PR #1092, opened 2017-05-27
Fix ackmate column numbers for multi-line matches

uhlissuh:
Currently, when the `--ackmate` flag is passed, `ag` prints incorrect column numbers for multi-line matches. The problem is that the `print_file_matches` function computes the column number for a match by subtracting the offset of the *most recent* line from the match's start offset. But in the case of a multi-line match, the match starts *before* the most recent line, so this subtraction produces a negative result.

Our solution was to carry an additional piece of state throughout the loop in `print_file_matches`: the offset of the line preceding the current match. This allows us to compute the column number correctly; we subtract the start offset of the line preceding the match from the match's start offset.

Paired with @maxbrunsfeld

### DIFF
--- a/src/print.c
+++ b/src/print.c
@@ -26,6 +26,7 @@ __thread struct print_context {
     size_t prev_line;
     size_t last_prev_line;
     size_t prev_line_offset;
+    size_t line_preceding_current_match_offset;
     size_t lines_since_last_match;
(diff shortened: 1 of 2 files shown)


ggreer:
Nice catch. Thanks for finding & fixing!

PR #1093, opened 2017-05-28
Run make with -r option (remove default rules).

ggreer:
Need to test this more & fully understand it before merging.


aswild:
Out of curiosity, why do this? It adds a layer of recursive make, and I assumed Automake handles all of the default rules correctly anyway.

Issue #535, opened 2014-11-22
Is it possible to add ag to MSYS2


ShuguangSun:
Is it possible to add ag to MSYS2? https://github.com/Alexpux/MSYS2-packages
It seems need a PKGBUILD, but I don't know how to do it.

oscarfv:
`ag` already is on MSYS2. It has some serious problems when running on the MSYS2 default shell, though, because `ag` behavior depends on if `stdin` is a pipe or a tty, and mintty makes `ag` think it is working on a tty. Use `--search-files`. Otherwise it works.

Swoorup:
@oscarfv Is it possible to fix this behaviour?

PR #1099, opened 2017-06-01
win32: Detect Cygwin/MSYS PTY

k-takata:
When running Win32 version of Ag on mintty or other Cygwin/MSYS terminal
emulators, Ag cannot detect Cygwin/MSYS pty and it causes some problems.
E.g.
https://github.com/ggreer/the_silver_searcher/issues/535#issuecomment-70944218
https://github.com/Alexpux/MSYS2-packages/pull/491

Import a check routine from https://github.com/k-takata/ptycheck .


avih:
@k-takata, I submitted #1146 without noticing this PR. One of its commits (`win32: detect and use normal ansi on msys2/cygwin tty`) seem to me pretty much identical to this one, but I think it's simpler.

Would you object if we go ahead with #1146 instead of this? Do you notice any functional differences other than the ones I mentioned at https://github.com/ggreer/the_silver_searcher/pull/1146#issuecomment-326061020 ?

avih:
@k-takata ping on my last question? any reason to prefer a solution with an order of magnitude more lines of code (and affected files and build procedure), which does basically the same as a much smaller solution?

k-takata:
I'll leave the decision to @ggreer .

avih:
But it's not only about this PR for ag. You use a variant of your fix in several other places in msys2 too. It's a big chunk of code which can really be much smaller and simpler. Don't you prefer smaller if it does the same thing?

k-takata:
Iscygpty is intended to be easily included in other software including non-MSYS2 software. The user should just include the two files into the project, add `#include "iscygpty.h"` and call `is_cygpty()` where needed.
The code should be configurable because it totally depends on the software whether support for WinXP is needed or not, or dynamic loading is needed or not. LOC is not important here.

avih:
> it totally depends on the software whether support for WinXP is needed or not, or dynamic loading is needed or not

This is partially off topic for ag, but Microsoft doesn't support XP anymore, and as far as I know neither do Msys2 nor cygwin. Therefore, jumping through hoops to support XP is, IMO, not required, because there's no current use case where a cygwin/msys2 tty exists on XP.

As for dynamic loading, it's a 5 lines function (after you remove XP support and the other redundant stuff). I don't think there's a good use case for dynamic loading. Note that the dynamic loading support is for DLLs which are used by the detection and are only required to enable the detection on XP. It's not for putting `cygpty` itself as a DLL. So if we ignore XP, there's no need for the dynamic loading code either.

And, specifically for ag, while you might want to provide XP support and other features when you use such patch downstream (e.g. in Msys2), imposing all of this on upstream is a lot of code for an unsupported use case - msys/cygwin on XP.

As a side note, and this is off topic for ag, I think the cygpty downstream msys2 patches (you mentioned vim and others) should also drop XP and simplify the code, because msys2 itself doesn't support XP anymore - again - as far as I know.

k-takata:
At least Vim still wants to support XP, so removing such configuration is not an option for iscygpty.

avih:
> At least Vim still wants to support XP, so removing such configuration is not an option for iscygpty.

I'm pretty sure the msys2 binary you build with the cygpty patch doesn't run on xp anyway. And even if it does, there's no value in the cygpty detection since msys2/cygwin don't run on XP.

Issue #797, opened 2015-11-25
Add information about MSYS2 to readme / wiki


elieux:
Somewhat related to #535.

MSYS2 provides both pre-built binaries and a build script and there was at least one person that was expecting to see that information here. If you're okay with adding the information to your readme and/or wiki, I can provide the texts.

Taverius:
I'll chime in here, as it seems the best place.

Following the instructions in [this gist](https://gist.github.com/k-takata/5124445)

I built `ag` from trunk just fine.

I've never used appveyor, but it looks like [MSYS2](http://www.msys2.org/) is supported, so you should be able to setup CI for windows there, and not redirect people to the existing outdated builds.

Or I could try to do that myself, and then send in a PR? I've never used appveyor before, so that might be a bit of an adventure, but CI for native windows builds sounds nice?

Frankly I'm surprised @k-takata hasn't already done so, given his work on compiling under MinGW, not to mention he wrote the instructions I used.

At the very least, you could add the instructions, its rather copy-pasta with a little waiting. for binaries to install themselves.

I can also update the instructions slightly, as the MSYS2 tools have evolved a little.

-----------------------

## For Win32/64 using [MSYS2](http://www.msys2.org/)

### Install Packages

Install the following packages using <code>pacman -S <i>package-name</i></code>:

* base-devel
* mingw-w64-{i686,x86\_64}-gcc
* mingw-w64-{i686,x86\_64}-pcre
* mingw-w64-{i686,x86\_64}-xz
* git

### Build The Silver Searcher

Open "MinGW-w64 Win32 Shell" or "MinGW-w64 Win64 Shell" from the start menu.

```
$ git clone https://github.com/ggreer/the_silver_searcher.git
$ cd the_silver_searcher/
$ ./build.sh PCRE_CFLAGS=-DPCRE_STATIC LDFLAGS=-static
$ strip ag.exe
```

Taverius:
Here we go, cleaned up and updated.

-----------------------

## For Win32/64 using [MSYS2](http://www.msys2.org/)

### Install Packages

Open "MSYS2 MSYS" from the Start menu.

Install the base-devel group, and git.

```sh
pacman -S base-devel
pacman -S git
```

Install the toolchain needed.

```sh
pacboy -S gcc:m
pacboy -S pcre:m
pacboy -S xz:m
```

Replace `:m` with `:x` to only install the x86_64 packages, and `:i` to only install the i386 packages.

[...]

PR #1100, opened 2017-06-03
Utilize AppVeyor

k-takata:
This PR adds scripts for running CI on [AppVeyor](https://www.appveyor.com/) (a CI service running on Windows).
(Related: #797)

This compiles Ag on the following environments:
* MSYS2 MinGW 64-bit
  Also create a zip package.
* MSYS2 MinGW 32-bit
  Also create a zip package.
* Cygwin (32-bit)
  Also run the test.

You can see the log at here: https://ci.appveyor.com/project/k-takata/the-silver-searcher/build/1.0.9

@ggreer 
To utilize this, you need to setup your appveyor account and enable this project.


k-takata:
The fix of #1104 is necessary for this.

PR #1106, opened 2017-06-17
Fix #896 - Stream decompress zipped files

### LINKED ISSUE #896: Implement zip file streaming
When searching large zip files (e.g. multi-GB logs or something), it's desirable not to decompress the whole thing in memory at one time.  Instead, a psuedo-FILE object can be created with `fopencookie(3)` (glibc) or `funopen(3)` (BSD) that only needs to seek forwards and decompress small chunks at a time, and that `FILE*` can be passed to the ordinary streaming search routine.

Here's an example of what `fopencookie`/`funopen`-wrapped zip streaming looks like: https://github.com/cemeyer/Zlib-FILE  (that implementation is permissively licensed, feel free to just drop it in to the_silver_searcher).

Aside from the FILE-wrapper code, the actual patch should be fairly small.  Something like this:

```
--- a/src/search.c
+++ b/src/search.c
@@ -315,14 +315,10 @@ void search_file(const char *file_full_path) {
     if (opts.search_zip_files) {
         ag_compression_type zip_type = is_zipped(buf, f_len);
         if (zip_type != AG_NO_COMPRESSION) {
-            int _buf_len = (int)f_len;
-            char *_buf = decompress(zip_type, buf, f_len, file_full_path, &_buf_len);
-            if (_buf == NULL || _buf_len == 0) {
-                log_err("Cannot decompress zipped file %s", file_full_path);
-                goto cleanup;
-            }
-            search_buf(_buf, _buf_len, file_full_path);
-            free(_buf);
+           log_debug("%s is a zip file. stream searching", file_full_path);
+           fp = decompress_file(zip_type, ...);
+           search_stream(fp, file_full_path);
+           fclose(fp);
             goto cleanup;
         }
     }
```

cemeyer:
Use the POSIX fopencookie(3) mechanism to produce a FILE object, and
then treat them the same as other non-mmapable streams (i.e., FIFOs).

Since some supported platforms do not support fopencookie(3) (Mac OS X,
maybe Cygwin, older BSDs), retain non-streaming zip file support.

Existing tests pass.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,7 @@
 ACLOCAL_AMFLAGS = ${ACLOCAL_FLAGS}
 
 bin_PROGRAMS = ag
-ag_SOURCES = src/ignore.c src/ignore.h src/log.c src/log.h src/options.c src/options.h src/print.c src/print_w32.c src/print.h src/scandir.c src/scandir.h src/search.c src/search.h src/lang.c src/lang.h src/util.c src/util.h src/decompress.c src/decompress.h src/uthash.h src/main.c
+ag_SOURCES = src/ignore.c src/ignore.h src/log.c src/log.h src/options.c src/options.h src/print.c src/print_w32.c src/print.h src/scandir.c src/scandir.h src/search.c src/search.h src/lang.c src/lang.h src/util.c src/util.h src/decompress.c src/decompress.h src/uthash.h src/main.c src/zfile.c
 ag_LDADD = ${PCRE_LIBS} ${LZMA_LIBS} ${ZLIB_LIBS} $(PTHREAD_LIBS)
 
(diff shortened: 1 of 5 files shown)


cemeyer:
For example, with a large compressed zero corpus (70 MB *compressed*), this makes completion actually possible:

Before patch:
```
$ ag -z foo test.out.xz
ERR: Unable to allocate 691011584 bytes to decompress file test.out.xz
ERR: Cannot decompress zipped file test.out.xz
```

After patch:
```
$ time ./ag -z foo test.out.xz
./ag -z foo test.out.xz  10.21s user 1.14s system 99% cpu 11.375 total
```

ggreer:
Wow, very nice!

PR #1107, opened 2017-06-17
Use Capsicum to drop privileges

cemeyer:
On platforms that support Capsicum (e.g., FreeBSD), drop privileges by
entering the capability sandbox.

Obstacles to entering the sandbox earlier include:
- popen of git in options parsing
- realpath in options parsing

All tests continue to pass on Capsicum-supporting platforms.


ggreer:
Neat. Could this break following symbolic links if those pointed outside the sandbox? (Ditto for following across filesystems, eg: directory containing mounts.)

cemeyer:
Yeah, symlinks that point above the current directory or to absolute paths are broken :-(.  We will have to manually follow symlinks with readlink+open.

Traversing across mountpoints works fine.  To the best of my knowledge, nothing in the Capsicum sandbox / VFS restricts to one filesystem at this time.

PR #1111, opened 2017-06-23
FreeBSD: Add support for setting CPU affinity

cemeyer:


ggreer:
Neat. Thanks!

cemeyer:
I was hoping to be able to generate the same `USE_CPU_SET` macro on FreeBSD instead of a second `#define`, but my m4/autotools skills weren't up to it.  Anyway, this seemed ok.

Thanks for merging :-).

PR #1118, opened 2017-07-05
languages added: asciidoc, org-mode and ipython-notebook

ojn:


ojn:
Forgot to add tests, will add again shortly

PR #1124, opened 2017-07-14
Add fusion as possible extension

mstruebing:
The neos-project(http://neos.io/) has it's own intermediate language called fusion which will be more and more dominant in NeosCMS projects (and maybe also Flow-Framework(the underlying framework)).

So it would be very nice if ag would be able to find contents in this files, I think it would be helpful for many people as the amount of developers for NeosCMS is growing more and more.

Thx!


mstruebing:
Any chance to get t his merged @ggreer ?

PR #1140, opened 2017-08-14
CMake support

osimola:
This is more invasive than usual language additions, since CMake has one special full filename: CMakeLists.txt. There are not many file formats use this naming scheme, Make could perhaps be one.

### DIFF
--- a/src/lang.c
+++ b/src/lang.c
@@ -5,130 +5,130 @@
 #include "util.h"
 
 lang_spec_t langs[] = {
-    { "actionscript", { "as", "mxml" } },
-    { "ada", { "ada", "adb", "ads" } },
-    { "asm", { "asm", "s" } },
-    { "batch", { "bat", "cmd" } },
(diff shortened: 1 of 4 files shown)


osimola:
There's a bug I didn't find last night, I'll reopen the PR after I have fixed that

adamryczkowski:
I just wanted to say, that there are people like me, who are really tired of typing `-G ".*\.cmake"` on each ag invocation, and implementing a shortcut `--cmake` for this would be awesome. :-)

adamryczkowski:
I am a C++ programmer myself. Perhaps I can help you with this bug? I've never contributed to the `ag` before.

osimola:
The fixed pull request is here: https://github.com/ggreer/the_silver_searcher/pull/1141. There are some older pull requests that attempt the same thing. I think it's safe to say at this point that either ag is unmaintained or the maintainer really doesn't want to support Make, CMake, Rake and so on. My solution was to switch to Ripgrep, which supports Make and CMake file types and is overall a better piece of software.

PR #1146, opened 2017-08-30
win32: native cygwin/msys2 tty support, and other minor improvements

avih:
All the commits in this PR are independent of the others, and I can split some of them to a new PR if you prefer. But they're all quite minor so I thought I'd start with a single PR.


avih:
Huh, enforced formatting. I'll handle it.

avih:
I've fixed the minor formatting issues, but now I don't see any travis indication (here at the PR page) of either success or failure. Not sure what's going on... but the formatting is hopefully fine now.

avih:
The last commit/force-push only updates the commit-message of `win32: detect and use normal ansi on msys2/cygwin tty` without touching any code.

k-takata:
Have you checked #1099?

avih:
> Have you checked #1099?

I did just now. It seems to do pretty much exactly the same thing[*] as my `win32: detect and use normal ansi on msys2/cygwin tty` commit, but in much more lines of code and files.

Their files also technically support XP while jumping through some hoops (which also require manual configuration of defines), but my code only "supports" XP by allowing to disable this detection on XP. Do note that XP support was dropped in msys2 and cygwin in 2016.

[*] There are very minor procedural differences I noticed:
-  When checking if we need to search stdin, my code aborts earlier if stdin is a tty, while their code leaves fstat and the other tests first, but it should behave identically as far as I can tell.
- Their pipe name matching/detection is a bit more elaborate and in quite more LOC. I wouldn't say one is more robust or better, as it's heuristics either way. I preferred a simpler matching, which is also more similar to what git-for-windows uses.

In general, it's pretty much the same as #1099, but in about 1/10 LOC of that PR.

k-takata:
> Also, I think #1099 has a potential buffer overflow here

Ah, you are right. I have updated my code. This is also used in Vim and some other software. So I sent patches for them.

src/decompress.h
```diff
@@ -19,7 +19,7 @@ ag_compression_type is_zipped(const void *buf, const int buf_len);
 
 void *decompress(const ag_compression_type zip_type, const void *buf, const int buf_len, const char *dir_full_path, int *new_buf_len);
 
-#if HAVE_FOPENCOOKIE
+#if HAVE_FOPENCOOKIE && HAVE_LZNA_H && HAVE_ZLIB_H
```
k-takata: s/LZNA/LZMA/

avih: Ouch :) Thanks! I'll push a fix.

avih:
Last force push fixes a type at the commit `build: workaround when fopencookie exists but zlib or lzma do not`. Thanks @k-takata for noticing it.

Issue #1147, opened 2017-09-01
compile failures when gzip or lzma support are disabled


radhermit:
Various places in src/zfile.c need to be #ifdef-ed and reworked so they don't assume support always is enabled. With gzip/lzma support disabled compiling fails with the following (just a snippet):

src/zfile.c:64:9: error: unknown type name ‘lzma_stream’
         lzma_stream lzma;
         ^~~~~~~~~~~
src/zfile.c: In function ‘zfile_read’:
src/zfile.c:210:5: error: unknown type name ‘lzma_ret’
     lzma_ret lzret;
     ^~~~~~~~
src/zfile.c:222:13: error: ‘LZMA_OK’ undeclared (first use in this function)
     lzret = LZMA_OK;
             ^~~~~~~
src/zfile.c:222:13: note: each undeclared identifier is reported only once for each function it appears in
src/zfile.c:73:89: error: request for member ‘next_out’ in something not a structure or union
 #define CNEXT_OUT(c) ((c)->ctype == AG_GZIP ? (c)->stream.gz.next_out : (c)->stream.lzma.next_out)
                                                                                         ^
src/zfile.c:231:16: note: in expansion of macro ‘CNEXT_OUT’
         while (CNEXT_OUT(cookie) >
                ^~~~~~~~~
src/zfile.c:73:89: error: request for member ‘next_out’ in something not a structure or union
 #define CNEXT_OUT(c) ((c)->ctype == AG_GZIP ? (c)->stream.gz.next_out : (c)->stream.lzma.next_out)

radhermit:
I forgot to mention, this is with 2.1.0.

sdlarsen:
Happens for me as well (Gentoo ~amd64). I too can confirm that enabling lzma makes it work. (I already had gzip support built in).

heroxbd:
I can reproduce this bug.

PR #1148, opened 2017-09-06
zfile: fix build and runtime errors

aswild:
As reported in #1147, the build fails if either `zlib.h` or `lzma.h` are unavailable.

Additionally, the feof check in zfile.c is backwards and causes a "truncated file" abort after searching the first zipped file.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,8 +1,12 @@
 ACLOCAL_AMFLAGS = ${ACLOCAL_FLAGS}
 
 bin_PROGRAMS = ag
-ag_SOURCES = src/ignore.c src/ignore.h src/log.c src/log.h src/options.c src/options.h src/print.c src/print_w32.c src/print.h src/scandir.c src/scandir.h src/search.c src/search.h src/lang.c src/lang.h src/util.c src/util.h src/decompress.c src/decompress.h src/uthash.h src/main.c src/zfile.c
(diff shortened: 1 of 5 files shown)


avih:
FWIW, I addressed this issue with a simple workaround at #1146 : https://github.com/ggreer/the_silver_searcher/pull/1146/commits/bf375de7b1177df7eb08eddc13c33c2e188c81c5 .

Your fix is clearly more complete, however, it does clutter the code quite a bit.

IMHO the best fix would be to refactor `zfile.c` such that it doesn't need gazilion `#ifdef`s. Or, alternatively, take the approach in my PR and just tie fopencookie to both lzma and zlib availability (and further remove the remaining `#ifdef`s - which I didn't do in my PR). There is a fallback anyway, so I think this approach is not too terrible.

Code with so many `#ifdef`s is not fun to maintain...

aswild:
Good point; I know the pain of dealing with code that's `#ifdef`'d to hell.

I can refactor `zfile.c` to put all of the zlib/lzma specific code into their own functions with a wrapper to check `cookie->ctype`, which would clean up `zfile_read` a lot.

aswild:
I tried adding a layer of abstraction (2 layers of inline functions, with stubs `#define`'d for unsupported modes), but it might be even worse than littering `zfile_read` with `#ifdef`s.

Perhaps your solution to just require both zlib and lzma to support fopencookie is better. (has anyone  benchmarked whether fopencookie even gives a noticeable performance boost?)

The alternate version is on my zfile-fix-redo branch: aswild/the_silver_searcher@24995eb239799e52ae09d47b7520d50a3b8b606c

Regardless of the above, I think we do need some form of my second commit to fix the `feof` handling, that missing `!` totally breaks zfile searching unless the file sizes happen to line up nicely. (aswild/the_silver_searcher@09b0bd8545d08d17ef70893fa15ae644e0bb8b2f)

PR #1150, opened 2017-09-10
Introduce an --invert-file-search-regex flag

### LINKED ISSUE #1138: Add --invert-file-search-regex flag
Just like `--invert-match` can be used to find words not matching a pattern, there should a flag that acts like the inverse of `--file-search-regex`, so that:

```shell
$ ag foobar --invert-file-search-regex baz
```

finds all files containing `foobar` whose filenames do *not* contain `baz`.

AlecBenzer:
Also contains a reformatting commit. Look at only the second commit for the functional diff.

Fixes #1138


AlecBenzer:
I don't seem to have permission to add a reviewer, but informal review request @ggreer :)

AlecBenzer:
Ping @ggreer

PR #1155, opened 2017-09-23
update readme to contain link to ack.vim

alykhantejani:


k-takata:
#1166 looks better than this.

Issue #972, opened 2016-09-22
ag does not honour --filename option when searching a single file


alexherbo2:

keith:
This is caused by the fact that the single file disable happens after the option parsing, specifically [here](https://github.com/ggreer/the_silver_searcher/blob/d10085cbccd602f15b6eca25880b17f5d20ab818/src/search.c#L541-L544).

keith:
I've submitted a PR with a fix for this issue https://github.com/ggreer/the_silver_searcher/pull/1161

PR #1161, opened 2017-10-01
Fix --filename with single file searches

keith:
Previously if you passed the --filename argument when searching a single
file, the value of print_path would be overwritten. This adds a new
PATH_PRINT_FORCED option for when users specifically pass --filename
(versus it being set by the default) so that it will not get
overwritten. I've also added a test to cover this case

Fixes https://github.com/ggreer/the_silver_searcher/issues/972

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -522,7 +522,7 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
                     opts.max_search_depth = atoi(optarg);
                     break;
                 } else if (strcmp(longopts[opt_index].name, "filename") == 0) {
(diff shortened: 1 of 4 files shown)


keith:
@ggreer this one should be ready if you have some time to take a look 🙏

keith:
@ggreer this one should be ready if you have some time to take a look 🙏

keith:
Rebased

keith:
@ggreer can you take a look at this small one?

keith:
@ggreer can you take a look at this small one?

keith:
I don't want this in my github.com/pulls list anymore so I'm going to close it

PR #1163, opened 2017-10-05
Added support for Idris, APL and J

Synthetica9:
Added support for the filetypes used by:

- [Idris](https://www.idris-lang.org) (`.idr`, `.lidr` and `.ipkg`)
- [APL]( https://en.wikipedia.org/wiki/APL_(programming_language) ) (`.apl`)
- [J](http://jsoftware.com) (`.ijs`)


ggreer:
Thanks!

PR #1166, opened 2017-10-11
Fix README.md link to ack.vim

me-and:


me-and:
It doesn't show up in the diff, but the current version of README.md looks like this in GitHub's rendering:

ack.vim not rendered as a link

me-and:
GitHub seem to have changed their rendering, so this is no longer necessary.

PR #1177, opened 2017-11-15
Update lang.c

kernhanda:
Adding support for *.tcc files for c++, powershell files and cmake files.


kernhanda:
I'll update the tests too.

kernhanda:
Looks like there's a fundamental problem with adding cmake support in that the filetypes only support extensions, not full filenames. Most CMake files are named `CMakeLists.txt`. I'll have to think about how to make this fix. 

Any suggestions are welcome.

PR #1184, opened 2017-12-01
Add --as-text option

murmour:
This PR adds a new `--as-text` option, which is equivalent to the `--text` option in GNU Grep.

Without this option, `ag` sometimes interprets cp1251-encoded sources as binaries. Effectively, the new option disables the binary detection heuristic, which makes perfect sense for situations where `ag` is asked to perform search only in files with a specific extension (corresponding to a particular language).

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -102,6 +102,7 @@ Search Options:\n\
                           uppercase characters (Enabled by default)\n\
      --search-binary      Search binary files for matches\n\
   -t --all-text           Search all text files (doesn't include hidden files)\n\
+     --as-text            Process binary files as if they were text\n\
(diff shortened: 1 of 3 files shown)


murmour:
Ping. I use this feature all the time to reliably grep through sources that contain non-ASCII characters.

proggga:
Hi, really required this param, to proper search

ethanabrooks:
Any follow up on this?

adamtaylor13:
Seconding this. I was surprised to find I couldn't just rename my binary file to `.txt` and ag through it :)

vandrlexay:
@murmour Thank you for MR, you saved me.

Also voting for inclusion this patch into main source tree. For some reason it treats regular text files in single byte encoding as binary ones.

PR #1190, opened 2018-01-01
Install bash completion to correct location

djmoch:
By default, bash completions are installed to
$dataroot/bash-completion/completions, so we should install there
instead.

Ref:
https://github.com/scop/bash-completion/blob/master/bash-completion-config.cmake.in


djmoch:
I don’t have a lot of experience with Travis CI, but the errors don’t appear to be related to my commit. If I’m wrong and I broke something, please let me know.

PR #1201, opened 2018-01-22
add conf/config as extension to search for

mstruebing:
This PR adds config to as a possible search extension.

I ofen want to grep for conf or config files, and this makes it easier.


mstruebing:
Any change to get this merged @ggreer ?

PR #1207, opened 2018-01-26
Add support for NO_COLOR environment variable

AbstractBeliefs:
When the NO_COLOR environment variable is present (with any value), disable
colour output. This can be overrident explicitly with the `--color` switch.

See also: http://no-color.org/

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -371,6 +371,11 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
         opts.stdout_inode = statbuf.st_ino;
     }
 
+    /* Suppress colors if user has NO_COLOR set (to any value). Allow --color to explicitly override */
+    if (getenv("NO_COLOR") != NULL) {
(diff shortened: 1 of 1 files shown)


AbstractBeliefs:
There's unfortunately no test for this: the test environment is has colour suppression due to the tty detection, and because of this all colour tests have the `--color` switch, overriding the environment variable.

A test could be implemented by removing the `--color` flag and disabling tty detection with another new flag, but I'm not sure this is a helpful path to go down.

ggreer:
I don't mean to annoy you by closing this. It's just that we seem to disagree on the usefulness of this feature. People can read my reasons here: https://news.ycombinator.com/item?id=17244313

Sorry.

kikonen:
I surely would have liked that also this tool would have simply conformed to NO_COLOR, instead of needing to extra tricks for this case :()

Thus 
```
alias ag="ag --nocolor"
```
it is

PR #1208, opened 2018-01-27
Add pug as a supported file type

jschneier:
Jade was renamed to Pug at some point because of a trademark issue. I noticed .jade is a supported file extension but .pug is not so I added it.


ggreer:
Thanks!

Issue #1199, opened 2018-01-16
Seg fault on large compressed files


wspurgin:
I've got a compressed log file `logs.gz`that is fairly large (3.49 GB) uncompressed:

Every time I run a pattern with `ag -z` on the compressed file, it seg faults right before getting to the ending byte mark (by a few KB). I can't include the full debug output, but I'll show the last few lines where it reports the seg fault.

```
DEBUG: Match found. File ../logs.gz, offset 3750991352 bytes.
DEBUG: Match found. File ../logs.gz, offset 3750999504 bytes.
DEBUG: Match found. File ../logs.gz, offset 3751014379 bytes.
DEBUG: Match found. File ../logs.gz, offset 3751014588 bytes.
DEBUG: Match found. File ../logs.gz, offset 3751017717 bytes.
DEBUG: Match found. File ../logs.gz, offset 3751037283 bytes.
Segmentation fault: 11
```

It reports getting to 3751037283 bytes (24,889 bytes short of the total 3751062172 bytes of the file) rather quickly, then hangs for a second, and finally dies with the seg fault.

I know this isn't a ton of information, and I'll gladly give more if someone can direct me on how to get more information (trace or otherwise) from `ag`.

You probably don't hear this enough, but I :heart: the silver searcher. Saves me on a daily basis from having to sit and wait for grep to finish ;)

Issue #1162, opened 2017-10-04
Cannot search large files: ERR: expected to read XXXX bytes but read 4294967295


koterpillar:
```
$ ls -l large-file
-rw-r--r-- 1 uuuuu ggggg 7775348191 Oct  4 11:06 large-file
$ ag foo large-file
ERR: expected to read 3480380895 bytes but read 4294967295
$ cat large-file | ag foo
(works)
$ ag --version
ag version 2.1.0

Features:
  +jit +lzma +zlib
```

This is macOS 10.12.6.

aswild:
ag isn't particularly optimized for searching huge files, especially when mmap is disabled (the [default on macOS](https://github.com/ggreer/the_silver_searcher/blob/master/src/options.c#L154)).

In [search_file](https://github.com/ggreer/the_silver_searcher/blob/master/src/search.c#L349) it will try to malloc and fill a buffer with the entire file contents, meaning the entire file is read into RAM before searching. This will be problematic if you're searching a 7 GB file and don't have that much free memory. Also it looks like `read()` on your OS might only read 2^32-1 bytes at a time.

Your workaround of `cat large-file | ag` (or `ag <large-file`) will enable stream searching which only buffers one line at a time. You could also try using the `--mmap` option.

Side note: size_t and off_t should be 64 bits if you're on a modern system, so I don't think they're overflowing, but the `%u` in the format string should really be `%lu` or `%zu` to handle values over 32-bits long when printing the error, which explains the "3480380895" in your error (7775348191 & 0xffffffff).

koterpillar:
I was trying to search in a directory with some large files present, and ag crashed as soon as it encountered one. Large files should either be searched using streaming or ignored with a warning.

iataylor:
So would a possible fix just be to give the user the option to have the program read in the file  (mimic the cat pipe) (if in OSX and if the file is determined to be too large)?

The other thing I noticed was that apparently there's a cutoff for an overly large file, would line by line search work for that as well, or would that be a performance hit?

cameron:
fwiw, the `--mmap` workaround is fine on os x, and ag is still noticeably faster than my old grep/find getup

pronoiac:
I'm just now realizing that it's an error, not a warning. So I've been bitten by this for a while. Any progress?

Issue #1157, opened 2017-09-26
Coredump when hitting a JAR with -z


wasamasa:
```
[...]
DEBUG: Found zip-based stream             
DEBUG: /home/wasa/code/stuff/Transportr/libs/gradle-witness.jar is a compressed file. stream searching   
ERR: Unsupported compression type: 3      
DEBUG: No match in /home/wasa/code/stuff/Transportr/LICENSE                         
DEBUG: pattern .gradle/ doesn't match path res/menu/recent_trip_select_actions.xml  
zsh: segmentation fault (core dumped)  ag -D -z foo ~/code/stuff/Transportr
```

Reverting to 2.0 resolves the issue. I'm on Arch and use the repository version.  Normally I'd try building from Git, but master seems to be the most recent branch and the last commit is the version bump to 2.1, so that seems to be a bit pointless.

Issue #1114, opened 2017-06-30
searching zip file not working


ri-aje:
I am on ubuntu 16.04 and got ag 2.0 from https://launchpad.net/ubuntu/artful/amd64/silversearcher-ag/2.0.0-1 and I have also tried git clone this repo and build locally (with or without --enable-libzip). either way, ag -z on zip files give me errors. such as:
`ERR: Unsupported compression type: 3
Segmentation fault (core dumped)`
and
`ERR: Zip files not yet supported: docs.zip
ERR: Cannot decompress zipped file docs.zip`

I checked issue history and found https://github.com/ggreer/the_silver_searcher/issues/743, and two PRs https://github.com/ggreer/the_silver_searcher/pull/843 and https://github.com/ggreer/the_silver_searcher/pull/855.

I have two questions:
1. does searching zip file work as of 2.0.0?
2. if so, does ag -z file.jar also work?
thanks.

PR #1221, opened 2018-02-27
Fixes overflows for large and compressed files

napcode:
Hi,

this PR probably fixes several issues in the github tracker related to search large (and compressed) files. Namely #1199, #1114 (maybe even #1157 and #1162). 
Please let me know if further changes are needed! 
Thanks for the beautiful silver searcher. 

-Mathias


wavexx:
I tested this for #1280 (with the supplied sample), but it didn't affect the results. Since you investigated a bit more into the issue, maybe there is some additional overflow that happens with that file?

PR #1225, opened 2018-03-09
Add -1 / --single-match option to only return the first result

MDUK0001:
Equivalent to ack's -1 option. Added for performance reasons in script that is only checking for the presence of a pattern.


MDUK0001:
Sorry about the last two commits - was an example commit but I forgot I had this pull request open!

Issue #206, opened 2013-05-01
-L (--files-without-matches) option does not work as expected.


vderyagin:
`ag -L 'foobarbaz'` just prints all filenames regardless of whether they contain match for a given search string or not.

Version 0.14.

jasonkarns:
This is still an issue in 0.32.0. can this be reopened?

blueyed:
@jasonkarns 
Confirmed.
It seems to list all files where the pattern does not match on any line.  E.g. `ag -L 'e|a|o|u|i'` will print much less.
I suggest creating a new issue for this.  Please link to here from it.

jasonkarns:
@blueyed #238 is a dupe of this and still open. Maybe best to take discussion there?

PR #1245, opened 2018-06-09
Fix -L again and decouple from -v

kpengboy:
This is a fix for #206/#238/#964/other duplicates, making the behavior of `ag -L` fully match `grep -L`.

This issue was previously fixed in PR #211; however, the patch applied there relied on a hack which no longer works correctly. Specifically, `-L` now reports a false positive whenever the query pattern appears only on the first or last lines of the text.

But there's also a more fundamental design error in `ag` that PR #211 doesn't address: `-L` is assumed to be equivalent to `-l -v`. But in reality, `-L`/`-l` and `-v` form two independent sets of flags. Here is the meaning of each combination of flags:

| | Without `-v` | With `-v` |
|---|---|---|
| **`-l`** | Print filename if any line matches pattern | Print filename if some line doesn't match pattern |
| **`-L`** | Print filename if no line matches pattern | Print filename if every line matches pattern |

The table above shows the behavior of `grep`, which (as it appears from the above issues) users expect `ag` to emulate. Currently `ag` implements `-l` correctly. Either `-L` or `-lv` also works, depending on what the file contents are. `-Lv` doesn't work at all.

This PR fixes the behavior of `-L` and decouples it from `-v`, fixing these issues.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -446,8 +446,9 @@ void parse_options(int argc, char **argv, char **base_paths[], char **paths[]) {
                 opts.casing = CASE_INSENSITIVE;
                 break;
             case 'L':
-                opts.invert_match = 1;
-            /* fall through */
+                opts.print_nonmatching_files = 1;
+                opts.print_path = PATH_PRINT_TOP;
+                break;
             case 'l':
                 needs_query = 0;
                 opts.print_filename_only = 1;
--- a/src/options.h
+++ b/src/options.h
@@ -60,6 +60,7 @@ typedef struct {
     int print_break;
     int print_count;
     int print_filename_only;
+    int print_nonmatching_files;
     int print_path;
     int print_all_paths;
     int print_line_numbers;
--- a/src/search.c
+++ b/src/search.c
@@ -2,8 +2,9 @@
 #include "print.h"
 #include "scandir.h"
 
-void search_buf(const char *buf, const size_t buf_len,
-                const char *dir_full_path) {
+/* Returns: -1 if skipped, otherwise # of matches */
+ssize_t search_buf(const char *buf, const size_t buf_len,
+                   const char *dir_full_path) {
     int binary = -1; /* 1 = yes, 0 = no, -1 = don't know */
     size_t buf_offset = 0;
 
@@ -13,7 +14,7 @@ void search_buf(const char *buf, const size_t buf_len,
         binary = is_binary((const void *)buf, buf_len);
         if (binary) {
             log_debug("File %s is binary. Skipping...", dir_full_path);
-            return;
+            return -1;
         }
     }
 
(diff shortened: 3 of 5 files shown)


kpengboy:
Description substantially rewritten after I reread PR #211

GilmarLira:
Just faced this issue today, and spend hours under the assumption I was not enough of a sophisticated  user. How can I help push this PR forward, @kpengboy, @ggreer? 

I’m using `ag` more and more at Carta, to track the adoption of our in-house React component library. Fantastic tool.

TrySound:
Ping guys. The feature is unusable.

agbodike:
@ggreer Any thoughts on this PR? It purports to resolve a number of outstanding issues.

ggreer:
Thanks for fixing this!

stronk7:
Maybe I'm being too bold... but just this issue, per se, deserves a release, IMO. It has been some good months since it was merged and it's not available in any distribution yet.

And without the fix... you really cannot trust ag results at all, potentially leading to wrong and/or dangerous actions if you aren't aware of this issue.

Just IMO, I've personally faced this issue at very least a couple of times in the past and, while the "mess" was detected on time and we did fallback to grep... current status isn't good at all!

Ciao :-)

ad-si:
Totally agree @stronk7, this is a really serious issue. A new release would be great @ggreer 😅

seveneves:
I am glad that there is a fix however I had to install the tool from sources to get the fixed version

mminot-yseop:
> Maybe I'm being too bold... but just this issue, per se, deserves a release, IMO. It has been some good months since it was merged and it's not available in any distribution yet.

I concur. I installed `ag` 2.2.0 from the Ubuntu 20 repositories, and I thought I was going crazy when I ended up with files being reported by _both_ `--files-with-matches` _and_ `--files-without-matches`.

silasb:
Woah, super confused as why this hasn't been released yet.  What's holding it back and how can we help?

adamtaylor13:
Yeah, I just banged my head for about 30 minutes trying to resolve this. Then I came and found it's actually not a PEBCAK error! Happy to do anything I can to wrap this up and get it released! @ggreer

hrvoj3e:
Will this fix ever be released?

`grep -L` works as it should but `ag -L` does not work at all.
 
How can I say to `ag` : _give me all files that don't have this string/pattern inside_ ?
Sorry, I use ag every day and forget this bug and waste time every time when using `-L` - this is so frustrating.

It would be great to see a new release so that Ubuntu and Arch can pick it up.

mminot-yseop:
> I use ag every day and forget this bug and waste time every time

The first time frustrated me enough for me to clone the repository and build the fixed version on my machine. Added the resulting `ag` binary to my `PATH` and voilà – didn’t even need to run install scripts or whatever. If you’re not too reticent about that, this can save a lot of headache in the long run.

> It would be great to see a new release so that Ubuntu and Arch can pick it up.

I sure hope so as well.

PR #1253, opened 2018-07-04
Adds stata files as a filetype option

bapfeld:
For dealing with Stata .do or .ado files


jschpp:
Yeah no. Check the whitespace in Line 320 https://github.com/ggreer/the_silver_searcher/pull/1253/files#diff-ed3c0db19eb051dfd596061720fd7b54R320

bapfeld:
Ah, got it - thanks @jschpp! Didn't realize I needed the leading spaces but we should be good to go now.

Issue #1278, opened 2018-10-31
Ag Ignore the ISO-8859 C program text, with CRLF line terminator, which is treated as Binary file


zhangxianyi:
Ag treat the CRLF file as binary. and ignore it.
but Ack works!
see the picture below:
![1](https://user-images.githubusercontent.com/6967453/47782622-9fe20980-dd3b-11e8-9dd9-c0a4f5e68a68.png)

jacwah:
Could you upload the failing file? I can take a look at it.

mljli:
[example.txt](https://github.com/ggreer/the_silver_searcher/files/2627618/example.txt)
@jacwah Uploaded.

```
$ file example.txt
***/example.txt  ISO-8859 text
$ grep Room example.txt
    m_shRoomID = -1;
    m_pRoom = NULL;
$ ag Room example.txt
// nothing
$ ag -a Room example.txt
Binary file ***/example.txt matches.
```

jacwah:
The test is in the `is_binary` function. It looks like it was written to only handle ASCII and UTF-8 encodings. CRLF is not the problem. ISO-8859 text may work depending on the ratio of English characters to non-English characters. Better support for other encodings should probably be treated as a feature request rather than a bug report.

zhangxianyi:
thanks all of you. 
the following issue give me the hint of solution.
[Support Japanese encoding (EUC-JP and Shift-JIS) #1288] : https://github.com/ggreer/the_silver_searcher/pull/1288
I modified the function `is_binary` to support the GBK code. it works
can I post the pull request?

ulidtko:
Well, in my opinion, that's not a bug of `ag`!

You got the output you searched for, right?

> Binary file **test/main.c matches**

--------
It's too naïve to hope that all software will change to accommodate broken files in drunk encodings.

*Millions* of other programs will start seeing your file as text rather than binary — once you fix the file. Which is damn easy to do:
 * convert your file [to UTF-8][•] (`iconv`)
 * stop using DOS-era line endings (`dos2unix`)

To configure your Windows machines to interoperate properly, search guides online. Sorry.

[•]: https://utf8everywhere.org/

PR #1288, opened 2018-12-01
Support Japanese encoding (EUC-JP and Shift-JIS)

zawataki:


zawataki:
@ggreer Could you please review this PR?

k-takata:
I think ag should accept any bytes over 0x7f to support any 8-bit encodings.
See also #1278.

zawataki:
@k-takata Thank you for comment and sorry for my late reply. But I don't know what a part of my changes is bad, so could you please teach me more details?

k-takata:
I think that current UTF-8 detection and your SJIS and EUC-JP detection should be all removed and `buf_c[i] >= 0x80` should be accepted.

zawataki:
@k-takata I understand, thank you for explanation. I will try to fix it.

PR #1289, opened 2018-12-03
add support in fortran types for preprocessed files with uppercases

aitzkora:
just had fortran files with uppercases extension to enable also preprocessed files


aitzkora:
the number of extensions was upper than 12. I remove some, I will propose another

PR #1312, opened 2019-03-12
Add --txt option

### LINKED ISSUE #1216: Interest in a `--txt` option
Avid `ag` user here. Wishing there was an easy way to search all `*.txt` files in a tree -- something clearly different from `--all-text`. Would there be any interest in this feature?

hawkins:
Intended to fix #1216

### DIFF
--- a/src/lang.c
+++ b/src/lang.c
@@ -115,6 +115,7 @@ lang_spec_t langs[] = {
     { "tcl", { "tcl", "itcl", "itk" } },
     { "terraform", { "tf", "tfvars" } },
     { "tex", { "tex", "cls", "sty" } },
+    { "txt", { "txt" } },
     { "thrift", { "thrift" } },
     { "tla", { "tla" } },
     { "tt", { "tt", "tt2", "ttml" } },
--- a/tests/list_file_types.t
+++ b/tests/list_file_types.t
@@ -338,6 +338,9 @@ Language types are output:
   
     --thrift
         .thrift
+    
+    --txt
+        .txt
   
     --tla
         .tla


JFLarvoire:
Thanks for submitting this patch.

You included just the txt extension. Could you also add other common text file extensions?  
For example the six listed there: https://en.wikipedia.org/wiki/List_of_file_formats#Text_files  

And also as many as possible from the common text-based formats from this list:  
https://en.wikipedia.org/wiki/List_of_file_formats#Text-based  
(With first and formost the .md extension used for all documentation on GitHub!)

aitzkora:
@hawkins : thanks!

hawkins:
@JFLarvoire I could, but all of those are already covered and imho unrelated...

- `.asc` etc. is already under `--asciidoc`,
- `.log` is under `--log`,
- and finally`.cfg` is so unlike what is normally in my .txt files that I'm not sure it's worth merging them

And the text-based ones are all over the place - just because files have plain text in them doesn't mean they're a txt file. I don't like the idea of seeing json, yaml, etc files when I specify that I wanted to see `.txt` files. Could you explain your reasoning for wanting all of these under one `--txt` flag please, @JFLarvoire ?

JFLarvoire:
The idea was to make it simple to search something in all non-source-code text files.
That's why `.md` immediately comes to mind. Markdown is just plain text, with some extra conventions on what punctuation means.
I mentioned `.log`, which I often search in, because I was unaware of the existing `--log`. Shame on me :-(
As for the other ones, I had no particular need to search in them. I just gave a pointer to these Wikipedia lists to show that other common text file formats existed, and might be worth including. (Even if they already appear in some other list, like `.asc`!)

proggga:
same as #1184

long-git-hash:
I would like this to be merged also.

PR #1324, opened 2019-06-20
Minor portability fixes

michaelforney:


michaelforney:
Not sure why the CI failed... looks like it couldn't access llvm.org?

Issue #1294, opened 2018-12-20
Why is AG_ZIP "Unsupported compression type" ? 


ampresent:
I get this error when I'm trying to search a .zip file
`ERR: Unsupported compression type: 3`

https://github.com/ggreer/the_silver_searcher/blob/1b06a9fd8d9ad0aaf235b4a752e605dfd4bb7ced/src/zfile.c#L87-L119

genivia-inc:
I'm surprised that type 3 is still used? Please note that *"Methods 1-6 are legacy algorithms and are no longer recommended for use when compressing files."* See the latest zip [APPNOTE 6.3.6](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT).

Given this, it seems unnecessary to support compression type 3.

For a wide range of use cases, compression types 0 (stored), 8 (deflate), 14 (LZMA). Perhaps also types 9 (enhanced deflate) and 95 (XZ). See also [wikipedia ZIP compression](http://fileformats.archiveteam.org/wiki/ZIP).  Note that type 0 is often used to store empty and small files and should not be flagged as an error by ag either.

I should add that [ugrep supports compression types 0, 8, 14, and 95](https://github.com/Genivia/ugrep#archives).

ampresent:
Thanks

PR #1333, opened 2019-09-01
Fix unsupported compression type message/segfault on ZIP and LZW

andreasstieger:
Fix unsupported compression type message/segfault on ZIP and LZW

is_zipped() had code for ZIP and LZW, but the streaming decompression
code did not cover that case. Remove.


andreasstieger:
This fixes #1294 and #1157.

PR #1353, opened 2019-12-27
To be able to build with lzma.h missing or zlib.h missing or both mis…

thebaptiste:
…sing

Even when passing the options --disable-lzma and/or --disable-zlib, the build (calling build.sh with the options) fails if lzma.h and/or zlib.h is missing because of undefined stuff in zfile_read

I hope this pull request will be helpful, even if you don't take it as it is

Regards


j-xella:
please incorporate this or similar fix into the main tree. I had problems building ag on my platform because of this.

PR #1377, opened 2020-04-15
Fix multiple global symbols definitions.

shlomif:
See the use of extern here:

* https://www.geeksforgeeks.org/understanding-extern-keyword-in-c/

* https://en.wikipedia.org/wiki/External_variable

*
https://stackoverflow.com/questions/496448/how-to-correctly-use-the-extern-keyword-in-c

This fixes GCC build on fedora 32/33 x64.

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -20,6 +20,8 @@
 const int fnmatch_flags = FNM_PATHNAME;
 #endif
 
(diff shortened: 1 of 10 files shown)


michaelforney:
I think this is a dup of #1324.

shlomif:
Thanks @ggreer !

PR #1389, opened 2020-06-22
options: fix segfault when TERM or HOME isn't set

aswild:
Regression from 3289ab8fba2141211487ef0572c995b122b5ad7b, if TERM isn't
set in the environment, getenv returns NULL which we shouldn't strcmp.
Similarly guard against using home_dir if it's null when looking for
global gitignore files.

This can be reproduced and tested with `env -i ./ag foo`.


ggreer:
Oops. Thanks for adding that test to prevent another regression.

PR #1410, opened 2020-09-29
Install shell completion files to correct locations

### LINKED ISSUE #271: bash completion scripts gets installed in the wrong place
The bash completion script currently gets installed into the wrong place. Bash completion scripts should always go to $(PREFIX)/share/bash-completion/completions regardless if bash completion is actually installed or not.

Currently the bash completion goes to $(PREFIX)/share/the_silver_searcher/.

carlwgeorge:
Bash completion files should be installed to `/usr/share/bash-completion/completions`, with a file name matching the command name.  Zsh completion files should be installed to `/usr/share/zsh/site-functions`, with a file name matching the command name prefixed with an underscore.

Resolves #271

https://github.com/scop/bash-completion/blob/master/README.md#faq
http://zsh.sourceforge.net/Doc/Release/Completion-System.html


okapia:
For the zsh completion, `_the_silver_searcher` is the correct naming.

In zsh, the naming typically follows that of rpm/deb package naming rather than the command names. This is partly because functions often handle multiple commands so, for example, `_subversion` handles completion for `svn` and `svn-admin` and `_ansible` handles all of `ansible-playbook`, `ansible-inventory` etc. Also note that in zsh, a completion function can handle other contexts like environment variable values which would not fit
a command-name-only naming scheme. So for example, `_ansible` also handles completion after an `ANSIBLE_STDOUT_CALLBACK=` assignment.

I can't find anything in the zsh documentation chapter that you reference to back up what you state. If there is something there that I missed, let me know and I'll arrange to fix it.

Note that it was me that contributed the original zsh completion function. There's also PR #1145 to update it which has not been merged. By now further updates are likely applicable.

carlwgeorge:
I'm not sure what I saw on that page back in September of 2020 that led me to use that as a reference.  Most other packages in Fedora seem to follow the convention of that file being named after the command they apply to, not the package name.  I understand your point about handling multiple commands, but in this case there is only one command, so why does it matter?  Over email you suggested to me that the point was to have the file match the package name, but I was already considering renaming the Fedora package from the_silver_searcher to ag.

PR #1412, opened 2020-09-30
Update RHEL/CentOS installation instructions

carlwgeorge:
This updates the RHEL installation instructions to resolve a number of issues.  Most importantly, the command didn't work before unless you already had epel-release installed.  It has to be installed before yum can see the_silver_searcher, so it needs to be a separate command.  Another important detail is that epel-release is only available in the default repositories of CentOS, not RHEL, so installing it by name doesn't work out the box.  A better approach is to use the convenience "latest" link provided by the EPEL project.  Finally, clarify that this works on both RHEL and CentOS, and for version 6 and up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,9 +62,10 @@ or
 * Fedora 22+
 
         dnf install the_silver_searcher
-* RHEL7+
+* RHEL/CentOS 6+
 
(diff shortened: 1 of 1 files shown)


carlwgeorge:
Any chances of getting this simple doc fix merged?

PR #1435, opened 2021-02-18
Better support for ignore inversions

immerrr:
This PR changes the approach to ignore file handling.

The current approach is 100% oriented towards performance, but unfortunately it loses very important pieces of "negate" functionality, because negations were applied only once and only if other ignore rules didn't match. This goes against the [gitignore spec](https://git-scm.com/docs/gitignore) where the negation lines are specifically designed to override all preceding rules.

There's a test attached to this change which describes a use case that was previously not working as it should. With this PR the test passes:
```
$ cat .gitignore
/a/*
!a/printme.txt
$ cat a/printme.txt
blah1
$ cat a/dontprintme.txt
blah2
$ ag blah
a/printme.txt:1:blah1
```

It is still not perfect, e.g. the example above won't work if the negation rule starts with a slash, `!/a/printme.txt`, because I was not sure about the right way to do this. Regardless, I hope it is enough to get the ball rolling and arrive to the final version eventually.

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -205,6 +205,19 @@ static int ackmate_dir_match(const char *dir_name) {
     return pcre_exec(opts.ackmate_dir_filter, NULL, dir_name, strlen(dir_name), 0, 0, NULL, 0);
 }
 
+/* Return 1 if the file is unignored by invert_regexes, 0 otherwise. */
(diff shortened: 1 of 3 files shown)


mcourteaux:
This needs merging.

canvural:
> This needs merging.

This project rarely receives updates now. So do not hope much.

PR #1443, opened 2021-04-14
Convert ANSI srting to utf-8 in arguments.

mattn:
In non-English locales, we often search files with multi-byte strings. But Windows does not convert strings. If you prefer the way to use manifest file, I'll update this PR to switch the way.

```xml
    <windowsSettings>
      <activeCodePage xmlns="http://schemas.microsoft.com/SMI/2019/WindowsSettings">UTF-8</activeCodePage>
    </windowsSettings>
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -43,6 +43,26 @@ int main(int argc, char **argv) {
     int workers_len;
     int num_cores;
 
+#ifdef _WIN32
+    UINT cp = GetACP();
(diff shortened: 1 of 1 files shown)


k-takata:
Just FYI, I have enabled the UTF-8 mode in [my Win32 builds](https://github.com/k-takata/the_silver_searcher-win32) using the following patch:
https://github.com/k-takata/the_silver_searcher-win32/blob/master/patches/0005-win32-Enable-UTF-8-mode-on-Windows-10-1903-or-later.patch

PR #1495, opened 2022-06-26
bash completion: do not expand tilde

maxnikulin:
Make behavior of bash completion for `ag` consistent with `ls`, `man`, etc. in respect to tilde expansion to home directory path. Do not replace `~` by `/home/user` in response to `[Tab]` in the following cases:

    ag pattern ~us
    ag pattern ~/src

Completion of file names is still handled by `_filedir`. As to other utilities, e.g. `ls` uses `_longopt` function that does not call
`_expand` at all, while `man` has call of `_expand` but it is bypassed for file names.

Expansion of `~` is controlled by the following setting in `inputrc`:

    set expand-tilde on

For details see [info "(bash) Readline Init File Syntax"](https://www.gnu.org/software/bash/manual/html_node/Readline-Init-File-Syntax.html). Alternatively `M-&` (`tilde-expand` command) may be used to expand tilde at particular point [info "(bash) Miscellaneous Commands"](https://www.gnu.org/software/bash/manual/html_node/Miscellaneous-Commands.html).

If you are going to revert this commit, please, add a comment with an example when `_filedir` is not enough and `_expand` is really necessary.

### DIFF
--- a/ag.bashcomp.sh
+++ b/ag.bashcomp.sh
@@ -6,8 +6,6 @@ _ag() {
   cur=$(_get_cword "=")
   prev="${COMP_WORDS[COMP_CWORD-1]}"
 
-  _expand || return 0
-
   lngopt='
     --ackmate
     --ackmate-dir-filter


maxnikulin:
Feel free to decline this pull request if I missed some use case.

My story is the following. I do not like when `~` is expanded by `/home/user` since it makes commands longer and more noisy. For a decade I had a hack from [1] redefining `_expand` in my `.bashrc` file. Recently I decided to try if the problem has been fixed and the snippet may be removed. For some period of time I was rather happy since everything worked as I prefer, but I have discovered that for `ag` tilde is still expanded. I have dug deeper and have followed links to other Ubuntu and Debian bugs mentioned in [1]. Some patch was applied and reverted [2]. I do not find the argument related to `scp` convincing, from my point of view it is `scp` completion that should be fixed in such case. On the other hand, since behavior of most commands is consistent with my expectations, I have no reason to complain that implementation of `_expand` should be patched. As a result I think that bash completion for `ag` needs a fix.

[1] [lp:622403 bash-completion: readline expand-tilde not acknowledged](https://bugs.launchpad.net/bugs/622403)
[2] [Ville Skyttä \[Bash-completion-devel\] `_expand`, scp and quoting](https://alioth-lists-archive.debian.net/pipermail/bash-completion-devel/2009-February/000973.html)

PR #1507, opened 2023-01-06
add lf file type for Lingua Franca programs

cmnrd:
See https://www.lf-lang.org/ and https://github.com/lf-lang/lingua-franca


lhstrh:
@ggreer it would be great if this could get merged. It's just a straightforward language addition, so this seems like a no-brainer to me. Thanks tor taking a look!

Issue #1531, opened 2023-09-19
memory leak in function 'parse_options'


JoraGevorgyan:
The memory leak is reported by a static analyzer tool developed at [CAST](https://www.linkedin.com/company/cast-center)
the [allocated memory](https://github.com/ggreer/the_silver_searcher/blob/a61f1780b64266587e7bc30f0f5f71c6cca97c0f/src/options.c#L340) will left without free when the function fails to [allocated another one](https://github.com/ggreer/the_silver_searcher/blob/a61f1780b64266587e7bc30f0f5f71c6cca97c0f/src/options.c#L342), because in function 'ag_malloc' it checks if memory is not allocated and exits whole program

ulzu:
Leaks are of no importance for short running programs. They can actually improve performance as the OS reclaims the resources anyways.

PR #1535, opened 2023-12-26
changed the way malloc is handled in options.c

DobryPpz:
Based on issue #1531 the function 'parse_options' is handling allocated memory incorrectly when there's a new allocation that failed. A list of so-far allocated memory is kept and automatically free'd when new allocation fails.


DobryPpz:
Based on issue https://github.com/ggreer/the_silver_searcher/issues/1531 the function 'parse_options' is handling allocated memory incorrectly when there's a new allocation that failed. A list of so-far allocated memory is kept and automatically free'd when new allocation fails.

